PR #2, opened 2024-11-13
Sqlite Wasm build: build script, README, shell polyfill

danleh:


danleh:
Regarding test duration/workload size: There is a convenient `--size` option for the `speedtest1` program, which I know set to 20 (default: 100), such that the benchmark takes ~600ms now. That would be acceptable for me (and lowering it even further would make me uneasy whether each individual sub-item still gets enough time and weather that gives realistic results).

Regarding noise/variance: Running the benchmark 20 times with the reduced workload size in `d8`, `jsc`, and SpiderMonkey's `js`, I get the following results:
| engine   |   mean(s) |        std |   std(%) |
|:---------|----------:|-----------:|---------:|
| v8       |    0.6621 | 0.00734059 |  1.10868 |
| jsc      |    0.7343 | 0.00593917 |  0.80882 |
| js       |    0.8056 | 0.0278292  |  3.45447 |

1-3% standard deviation sounds fine to me.

kmiller68:
Nice! Looks good so far!

One thing that might be worthwhile is to put the harness/driver hooks in their own file. I noticed when I tried to recompile some of the existing tests we couldn't just drop in emcc's new `.js` files because the harness was mixed in.

danleh:
It now works in the CLI runner of JetStream and when running standalone via `sqlite3/runner.js` in shells. I still need to
- [ ] intercept one warning output/print, that originates internally in sqlite3, and
- [ ] make sure it works correctly (without OPFS) in the browser runner.

danleh:
@kmiller68 This is ready for your review from my side.

It's still using and scored with the "old" `WasmBenchmark` class, but we can eventually migrate it to the same scoring as you do for TSF-wasm in #18.

JetStreamDriver.js
```diff
@@ -1109,16 +1109,17 @@ class WasmBenchmark extends Benchmark {
             };
 
             oldPrint = globalObject.print;
+            oldConsoleLog = globalObject.console.log;
             globalObject.print = globalObject.printErr = (...args) => {
                 if (verbose)
-                    console.log('Intercepted print: ', ...args);
+                oldConsoleLog('Intercepted print: ', ...args);
```
kmiller68, line 1115: This indentation seems weird? Can we keep the old indentation?

cli.js
```diff
@@ -35,6 +35,8 @@ const isSpiderMonkey = typeof newGlobal !== "undefined";
 if (isSpiderMonkey)
     globalThis.readFile = readRelativeToScript;
 
+if (typeof arguments !== "undefined" && arguments.length > 0)
+    testList = arguments.slice();
```
kmiller68, line 39: Nice!

sqlite3/runner.js
```diff
@@ -0,0 +1,80 @@
...
+  const pVfs = capi.sqlite3_vfs_find(vfs);
+  if (!pVfs) {
+    console.error("Error: Unknown VFS:", vfs);
+    return;
+  }
+
+  // These arguments should match the upstream browser runner `speedtest1.html`.
+  let argv = [
+    "speedtest1",
+    "--singlethread",
```
kmiller68, line 52: I guess for webpages they probably only need single threaded? Does it work without single threaded on the web or is that argument de facto required?

kmiller68:
Overall looks good to me. Can we fix the one indentation issue first though?

kmiller68:
> This PR also currently includes/depends on https://github.com/WebKit/JetStream/pull/15.

Do you want to land that one first or do it here. I'm ok with either.

sqlite3/runner.js
```diff
@@ -0,0 +1,80 @@
+// Copyright 2024 the V8 project authors. All rights reserved.
```
kmiller68, line 1: It seems like the existing benchmarks use `benchmark.js` for their driver code. It's probably worth it to stay consistent.

danleh: Right, renamed.

sqlite3/runner.js
danleh: I copied this argument from the upstream `speedtest1.html` harness, which also uses `--singlethread` in the Web version of this benchmark. I don't see any observable difference with or without it, however. I have a minor preferences for leaving it as-is, to keep it consistent with the upstream benchmark harness. WDYT?

danleh:
> Overall looks good to me. Can we fix the one indentation issue first though?

Done.

Also renamed `runner.js` to `benchmark.js` as you proposed for consistency.

> > This PR also currently includes/depends on #15.
> 
> Do you want to land that one first or do it here. I'm ok with either.

Let's just land this here, I'll close #15 subsequently.

danleh:
Ready from my side (sorry again for the delay.)

kmiller68 (approved):
Looks good! Thanks.

PR #4, opened 2024-11-22
Fix and cleanup 3d-cube

camillobruni:
- Avoid accidental globals
- Add Q.LinePixels to avoid side-effect free loop
- Improve indendation a bit
- Remove unused variable
- Update bomb-workers version as well


kmiller68 (approved):
LGTM

PR #5, opened 2024-11-22
Display selected benchmark names by default

camillobruni:
By default all benchmark data is "censored".
Showing the benchmark names at the start is more helpful, especially when filtering is enabled.

Screenshot 2024-11-22 at 10 27 13

### DIFF
--- a/JetStream.css
+++ b/JetStream.css
@@ -301,6 +301,7 @@ a.button {
     background: linear-gradient(160deg, rgba(249,249,249,1) 0%, rgba(238,238,238,1) 100%);
     border-radius: 3px;
 }
+
 .benchmark h3 {
     color:  rgb(183, 183, 183);
 }
--- a/JetStream.css
+++ b/JetStream.css
@@ -296,10 +296,13 @@ a.button {
 
 }
 
-.benchmark h3, .benchmark h4, .benchmark .result, .benchmark label {
+ .benchmark h4, .benchmark .result, .benchmark label {
     color: transparent;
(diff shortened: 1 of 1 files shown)


danleh:
Makes sense to me.
Since I don't have the power to LGTM here, how shall reviewing proceed in the future? Can we add additional reviewers? How was it done in the Speedometer 3 effort?

JetStream.css
```diff
@@ -296,11 +296,14 @@ a.button {
 
 }
 
-.benchmark h3, .benchmark h4, .benchmark .result, .benchmark label {
+ .benchmark h4, .benchmark .result, .benchmark label {
```
kmiller68, line 299: There's an extra space here probably worth cleaning up for consistency.

kmiller68:
> Since I don't have the power to LGTM here, how shall reviewing proceed in the future? Can we add additional reviewers? How was it done in the Speedometer 3 effort?

Not sure how SP3 did it. My assumption is that we'd want a WebKit reviewer to review all the changes until we've agreed on an open governance policy though, I'll ask folks internally though. Apple has this whole week of for US Thanksgiving so I probably won't have an answer this week, unfortunately.

danleh:
Sure, makes sense! :) Let's discuss in the next meeting and hear what the WebKit/Apple folks are saying.

JetStream.css
camillobruni: 👍 
oh I should probably copy over the `npm run format` commands from Speedometer in another PR.

kmiller68: Still a space here.

camillobruni: ups, I'm blind to formatting :P

PR #7, opened 2024-11-22
Clean up variables

camillobruni:
- Use const where possible
- Use globalThis.allIsGood consistently

Drive-by-fix:
  - Add consol.error in html initialize function for better debugging


danleh:
LGTM (again, no power, let's discuss how to organize.)

kmiller68 (approved):
LGTM too.

kmiller68:
You don't need to do in this PR but we could probably replace all(?) `globalObject.foo`s with `globalThis.foo` too. I don't think `globalThis` was a thing when JS2 was written.

PR #8, opened 2024-11-22
Use regexp results

camillobruni:
Octane's regexp benchmark only access the length of the result string but never it's contents.
This seems slightly unrealistic to optimise just for the .length case when using `exec()`.


kmiller68 (approved):
Seems reasonable to me and I didn't hear any objections at the last sync.

PR #11, opened 2024-11-26
Implement end2end tests

camillobruni:
- Copy over package.json from Speedometer3
- Implement `npm test:chrome`, `npm test:firefox`, `npm test:safari`
- Add better support for custom `worstCaseCount` and `iterationCount` to lower test duration
- Pass along `iterations` to `Benchmark.validate` to fix unipoker expectations
- - Add `JetStreamReady` and `JetStreamDone` events to the `Driver` for easier testing


tests/server.mjs
```diff
@@ -0,0 +1,51 @@
...
+        const stream = fs.createReadStream(streamPath);
+        return { found, ext, stream };
+    };
+
+    const server = http
+        .createServer(async (req, res) => {
+            const file = await prepareFile(req.url);
+            const statusCode = file.found ? 200 : 404;
+            const mimeType = MIME_TYPES[file.ext] || MIME_TYPES.default;
+            res.writeHead(statusCode, { "Content-Type": mimeType });
```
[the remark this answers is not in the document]
camillobruni: done, and switched `npm run server` to use the custom test server as well now.

JetStreamDriver.js
```diff
@@ -27,29 +27,44 @@
...
     const urlParameters = new URLSearchParams(window.location.search);
     shouldReport = urlParameters.has('report') && urlParameters.get('report').toLowerCase() == 'true';
     if (urlParameters.has('test'))
         customTestList = urlParameters.getAll("test");
+    globalThis.testIterationCount = getIntParam(urlParameters, "iterationCount");
+    globalThis.testWorstCaseCount = getIntParam(urlParameters, "worstCaseCount");
 }
 
+
+
```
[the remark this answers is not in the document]
camillobruni: removed.

JetStreamDriver.js
```diff
@@ -54,17 +56,14 @@ if (typeof(URLSearchParams) !== "undefined") {
     const urlParameters = new URLSearchParams(window.location.search);
     shouldReport = urlParameters.has('report') && urlParameters.get('report').toLowerCase() == 'true';
     if (shouldReport)
-        startDelay = 4000;
-    if (urlParameters.has('startDelay'))
-        startDelay = urlParameters.get('startDelay');
+        globalThis.startDelay = 4000;
     if (urlParameters.has('test'))
         customTestList = urlParameters.getAll("test");
+    globalThis.startDelay = getIntParam(urlParameters, "startDelay");
```
kmiller68, line 62: This seems like it will always set `startDelay` to` undefined` when the there's no `startDelay` URL parameter. I think that's fine for `testIterationCount` and `testWorstCaseCount` since they'd always be undefined on the web as it is. Although maybe it's worth changing there too so folks aren't surprised?

camillobruni: I hoped to make the initialization explicit by the lines above `globalThis.startDelay ??= undefined;` at least for me that's a bit more consistent across all configuration params.

Medium term I'd like to adapt the Params object from Speedometer (this also allows us to easily copy over the ?developerMode UI).

Maybe I missed your point here, but when running via the shell we now also always initialize startDelay=undefined.
With the html runner we'd initialize to undefined by default (and set it again to undefined if there is no matching url param)

But generally not feeling strongly about this one :)

camillobruni:
I've now merged in the latest PR from speedomter to use the local-web-server package so we don't have to write our own little webserver.

JetStreamDriver.js
kmiller68: I think the issue is when you do `index.html?report=true` `startDelay` ends up as `undefined` rather than `4000`. If you moved the report check below this and changed it to `if (shouldReport) globalThis.startDelay ??= 4000` I think that would fix it.

Safari, at least, relies on the startDelay to make sure the browser has calmed down after launching enough to prevent noise.

camillobruni: ok, then I understood correctly.
Keeping the old behavior with `report=true` sg.

kmiller68 (approved):
LGTM

PR #13, opened 2024-11-26
Directly use BenchmarkClass

camillobruni:
- Avoid the `plan.benchmarkClass` and directly use the dedicated `BenchmarkClass` and get some free pre-validation.
- Pull out the `fetchResources()` call from the `BenchmarkClass` constructor


JetStreamDriver.js
```diff
@@ -689,7 +691,7 @@ class Benchmark {
 
     async doLoadBlob(resource) {
         let response;
-        let tries = 3;
+        const tries = 3;
         while (tries--) {
```
kmiller68, line 695: This seems wrong.

kmiller68 (commented):
Overall seems fine although I think you have a bug. This also needs another rebase (sorry).

JetStreamDriver.js
camillobruni: ups... definitely.

camillobruni:
just rebased (was easy enough this time :).

camillobruni:
resolved conflicts with latest main.

kmiller68 (approved):
LGTM

PR #14, opened 2024-11-26
Replace groups with more flexible tags

camillobruni:
Every Benchmark can now have multiple tags which will simplify classifying experimental benchmarks better in the future.
This currently depends on #13.


JetStreamDriver.js
```diff
@@ -570,16 +570,23 @@ class Benchmark {
...
         this.scripts = null;
-
         this._resourcesPromise = null;
     }
 
+    _validateTags(tags) {
+        const validated = new Set();
+        for (const tag of tags) {
+            if (!TAGS.has(tag))
+                throw new Error(`Invalid benchmark tag "${tag}"`);
```
kmiller68, line 584: Nit: Seems a bit unnecessary since tags are mostly a debugging feature. We also only use the `TAGS` Set here so IMO I would skip this and remove `TAGS`. If you make that change you could just turn `this.tags = this._validateTags(plan.tags);` above into `this.tags = Set.from(plan.tags);` and optionally add 
```
if (this.tags.size != plan.tags.length) 
    throw new Error(`Duplicate tags ${plan.tags.filter((tag, index) => plan.tags.indexOf(tag) !== index)} in plan of ${this.name}`);
```

JetStreamDriver.js
```diff
@@ -1870,7 +1879,7 @@ const BENCHMARKS = [
         ],
         iterations: 160,
         worstCaseCount: 16,
-        testGroup: BigIntMiscGroup,
+        tags: ["BigIntMisc",]
```
kmiller68, line 1882: Nit: some of these have trailing commas some don't.

kmiller68 (approved):
LGTM with some nits

JetStreamDriver.js
camillobruni: I was mostly focused on preventing typos, but this can be just a unittest in the end (I'll re-add that after the cli test PR has landed).

JetStreamDriver.js
```diff
@@ -2232,48 +2241,50 @@ for (const name of WTB_TESTS) {
...
         worstCaseCount: 1,
-        testGroup: WTBGroup
+        tags: ["WTB"]
     }));
 }
 
 
 const benchmarksByName = new Map();
-const benchmarksByGroup = new Map();
+const benchmarksByTag = new Map();
```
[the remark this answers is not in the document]
camillobruni: ah yeah, this was easy enough to fix.

JetStreamDriver.js
camillobruni: let me go with trailing commas everywhere!

JetStreamDriver.js
```diff
@@ -1733,68 +1732,68 @@ const BENCHMARKS = [
...
-        testGroup: SimpleGroup
+        tags: ["Simple"],
     }),
     new AsyncBenchmark({
         name: "doxbee-promise",
         files: [
             "./simple/doxbee-promise.js",
         ],
-        testGroup: SimpleGroup,
+        tags: ["Simple",],
```
kmiller68, line 1750: Ah my previous nit was about `["Simple"]` vs `["Simple",]` but it's not a huge deal.

JetStreamDriver.js
```diff
@@ -1448,26 +1445,28 @@ class WasmLegacyBenchmark extends Benchmark {
...
+    "BigIntNoble",
+    "BigIntMisc",
+    "Proxy",
+    "ClassFields",
+    "Generators",
+    "Wasm",
+    "WorkerTests",
+    "WSL",
+    "WTB"
+]);
```
kmiller68, line 1467: This seems unused now.

JetStreamDriver.js
camillobruni: ah this was easy to fix...

PR #15, opened 2024-11-26
Use arguments to CLI runner as list of items to run

danleh:


danleh:
Included as part of #2.

PR #16, opened 2024-11-27
Store times and scores explicitly

camillobruni:
Prepare the benchmarks for more flexible result calculation and visualisations:
- Store all metrics as times and as scores
- Add uiFriendlyScore to mark displayed values explicitly as score
- Rename Benchmark.subTimes to Benchmark.subScores


kmiller68:
What's the goal here? I think I'm missing some context.

camillobruni:
I want to make the units explicit and make sure we have time+score everywhere available for consistency.
Personally not a big fan of scores (especially if they're arbitrarily scaled), and we have `[ms]` everywhere in speedometer.

Preparing this should make it a bit easier in the future whether we want to switch to times and only have a global score or not.

JetStreamDriver.js
```diff
@@ -280,30 +284,30 @@ class Driver {
...
+            summaryElement.innerHTML = "<div class=\"score\">" + uiFriendlyScore(geomean(allScores)) + "</div><label>Score</label>";
             summaryElement.onclick = displayCategoryScores;
             if (showScoreDetails)
                 displayCategoryScores();
             statusElement.innerHTML = '';
         } else if (!dumpJSONResults) {
             console.log("\n");
             for (let [category, scores] of categoryScores)
-                console.log(`${category}: ${uiFriendlyNumber(geomean(scores))}`);
+                console.log(`${category}: ${uiFriendlyScore(geomean(scores))}`);
```
kmiller68, line 308: Can we do `${category}: ${uiFriendlyScore(geomean(scores))}, ${uiFriendlyDuration(geomean(times))}`?

kmiller68 (commented):
Ah, gotcha, I'm mostly indifferent on time vs score as mention previously. Maybe for now we could print the times too for CLIs?

JetStreamDriver.js
kmiller68: I guess we can do this in a follow up PR.

kmiller68 (approved):
LGTM

PR #18, opened 2024-11-27
Move wasm benchmarks to JS benchmark style scoring

kmiller68:
As discussed with other browsers the plan for JS3 is to score wasm the same way we score JS (with First Iteration, Average, and Worst N). Each iteration (other than the first) runs in about 60ms on my M4 Max MBP.

Move existing benchmarks to WasmLegacyBenchmark and add a new WasmEmccBenchmark that supports running iterations. It's currently an async benchmark subclass since the compile is async. 

As a proof of concept this patch converts tsf-wasm to use the new scoring. I roughly expect the process to add a C++ library that doesn't already have its own wasm build script to be: compile with `-s MODULARIZE=1 -s EXPORT_NAME=setupModule` then add a benchmark.js with something like 
```
class Benchmark {
    async runIteration() {
        if (!Module._runIteration)
            await setupModule(Module);

        Module._runIteration(150);
    }
};
```

I sampled the old test but recompiled and the new JS-style version of the test and the samples seemed quite similar. So it seems like this benchmark change doesn't radically alter the characteristics of at least this test.

Lastly, get emcc to ouput the .symbols object so we know which wasm function is which index. We could emit a name section into the binary but my assumption is that most production code does not ship with that so it might change the benchmark in an unrealistic way.


kmiller68:
See https://github.com/WebKit/JetStream/pull/18/commits/d1534f036ba747749e5dee5737c0ff01fe19cdcc for details on the time spent in each function breakdowns. Weirdly even though the test runs significantly longer JSC still spends the same amount of time in our BBQ (template JIT) tier. Maybe that's a bug on our end though?

kmiller68:
@camillobruni @danleh Do either of you have thoughts/comments/concerns? I would CC the Mozilla folks too but I don't have their GH handles...

danleh:
Sorry for the late reply. I'll take a look later today.

wasm/TSF/benchmark.js
```diff
@@ -0,0 +1,33 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+class Benchmark {
+    async runIteration() {
+        if (!Module._runIteration)
+            await setupModule(Module);
+
+        Module._runIteration(150);
```
danleh, line 31: So just to clarify the workload size before vs. now: It used to be the default of `count = 1000` because `main` was invoked without arguments, and now we are at `defaultIterationCount = 120` times `150` per `runIteration`, right?

It seems the line item is running quite a bit longer now cf. before (~12s vs. ~1.7s, or roughly 130B instructions vs. 15B on an x64 workstation). I see more functions using the top-tier now, which is fine by me. We just might have to lower it, in case the overall benchmark becomes too long to run.

wasm/TSF/build.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh
+
+set -euo pipefail
+
```
danleh, line 4: WDYT about adding some log file with a last build timestamp and Emscripten version for reproducibility? Something like
```
BUILD_LOG="$(realpath build.log)"
echo -e "Built on $(date --rfc-3339=seconds)\n" | tee "$BUILD_LOG"

echo "Toolchain versions" | tee -a "$BUILD_LOG"
emcc --version | head -n1 | tee -a "$BUILD_LOG"
```
(shameless copy from [sqlite3/build.sh](https://github.com/WebKit/JetStream/pull/2/files#diff-e7f726f906491117444ff9d200e03b79188411304fd391ec888a420f1d7ffbd2))

wasm/TSF/tsf.js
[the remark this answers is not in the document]
danleh: And just to clarify: You did rebuild with a more up-to-date Emscripten version, right? So some changes in, say, execution profile, could simply come from toolchain evolution, right?

JetStreamDriver.js
```diff
@@ -986,6 +1007,94 @@ class AsyncBenchmark extends DefaultBenchmark {
     }
 };
 
+// Meant for wasm benchmarks that are directly compiled with an emcc build script. It might not work for benchmarks built as
+// part of a larger project's build system or a wasm benchmark compiled from a language that doesn't compile with emcc.
+class WasmEMCCBenchmark extends AsyncBenchmark {
```
danleh, line 1012: LGTM for now, long term it would be nice to get rid of some of the code duplication, e.g., the `prerunCode` below is mostly (but not fully) equal to the one in `WasmLegacyBenchmark` right?

danleh:
> @camillobruni @danleh Do either of you have thoughts/comments/concerns? I would CC the Mozilla folks too but I don't have their GH handles...

Overall looks great to me, thanks!

In terms of Mozilla folks, I only know of (some of them mentioned their GitHub handles in the WebKit JetStream discussion Slack channel)
- Ryan Hunt @eqrion
- Ben Visness @bvisness
- Iain Ireland @iainireland
- Brian Grinstead @bgrins

wasm/TSF/benchmark.js
danleh: Excerpt of a bottom-up profile in d8 before:
image
vs. now:
image
I'm happy to land this, we should just be aware that this is now moving more into the direction of measuring peak performance. (Was this a conscious decision?)

JetStreamDriver.js
kmiller68: Yeah, I think the goal is to eventually just delete WasmLegacyBenchmark once we have all the benchmarks we care about in the JS style system. If we want to save them for posterity we can reevaluate.

wasm/TSF/build.sh
kmiller68: Yeah, I was thinking we should just have one script to do it rather than duplicating the logic. That said, I can do that in a follow up https://github.com/WebKit/JetStream/issues/22. I'll copy your logic for now.

wasm/TSF/benchmark.js
kmiller68: Yeah, I should probably reduce the iteration count. I did that in my followup patches for other benchmarks, I'll do it here too. That said, it seems like your samples include the harness code? That's probably not ideal.

The function numbers also changed for me when I recompiled the test, I ended up just simulating the old benchmark by running one iteration with count = 1000 in the samples I put in my commit message.

wasm/TSF/tsf.js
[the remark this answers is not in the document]
kmiller68: Yeah, the function numbers are completely different. Now we have the `.symbols` so at least you can manually map between emcc versions.

JetStreamDriver.js
```diff
@@ -36,11 +36,17 @@ globalThis.dumpJSONResults ??= false;
 globalThis.customTestList ??= [];
 
 let shouldReport = false;
+let startDelay;
```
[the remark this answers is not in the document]
kmiller68: Yeah, this should be independent. The delay is because our internal A/B testing harness restarts the browser between full benchmark runs to reduce variance or caching. So we want to wait a few seconds for the browser to to calm down before starting the test, at least, IIRC.

wasm/TSF/build.sh
kmiller68: It seems like `date --rfc-3339=seconds` isn't supported by MacOS's out of the box date. I'm gonna go with `date -u '+%Y-%m-%dT%H:%M:%SZ'`

kmiller68:
I'll wait until EoD to merge in case any Mozilla folks have comments .

wasm/TSF/benchmark.js
danleh: Correct, I just took a profile of the whole d8 execution with Linux perf (via the helper explained in https://v8.dev/docs/linux-perf#profiling-d8-with-linux-perf-d8.py which takes care of injecting the samples from JIT-ed code). If startup or benchmark harness take a significant portion of the overall profile, I would say the workload is too short/light (as it probably was before).

wasm/TSF/build.sh
danleh: Ah, good to know and keep in mind for an eventual (shared?) build script.

JetStreamDriver.js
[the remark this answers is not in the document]
camillobruni: All in on more configurable test params!

I'll probably copy the params.mjs from speedometer as well and then we have a more central place to store all the test parameters.

PR #19, opened 2024-11-29
Improve async-fs

camillobruni:
Async-FS is heavily dominated by Math.random (Maybe we should test the real Math.random calls in some other workload). 
This PR replaces the already mocked Math.random with a very simple counter + modulo approach.


kmiller68 (approved):
LGTM.

PR #21, opened 2024-12-19
Remove float-mm.c asm.js test

kmiller68:
asm.js sees very low usage in the wild now that wasm has been shipping for a while and largely replaced it. Thus we don't cosider it worth optimization effort to improve asm.js performance going forward.


danleh:
Agreed, removing asm.js workloads SGTM. Let's confirm with Mozilla folks (either here or in the next sync meeting).

kmiller68:
At the sync meeting folks agreed to remove this.

PR #23, opened 2024-12-21
Move tfjs to first, average, worst scoring.

kmiller68:
Since this test is relatively slow per iteration I reduced the iteration counts to: non-simd: 15 iterations with worst 2.
simd: 80 iterations with worst 4 (default).

Also, I moved all the benchmark files to a tfjs directory to help organize the wasm directory a bit.


danleh:
Overall this looks good to me, with some high-level comments though:
1. Unlike the other Emscripten workloads, this doesn't contain a build script. How are the Wasm and JS files generated? Did you compile them at some point, or do they come from NPM packages? Ideally, we should be able to recompile with an up-to-date toolchain, but alternatively can we at least download them from somewhere, in case the upstream packages gets updated? `in-depth.html` mentions several repositories for Tensorflow.js, the Wasm backend, and models, but I am not sure how this is setup or copied together.
2. Even with the lowered iteration counts, in particular the SIMD test is pretty slow (in the order of 30 to 40s). Could we reduce the workload per iteration or just lower `iterations` further, say to 30?
3. Looking at the profile/flamegraph, a substantial number of samples are spent on parsing the `.js` files and an even larger amount on `base64ToArrayBuffer`, which is used when reading the model weights. Other folks have reported JetStream 3 initialization taking much longer than before. Could this be because of the large `.js` files and in particular model weights of this benchmark? Should we choose maybe just one or two models/inference tasks, not four?

kmiller68:
r.e. your comments:

1) Yeah setting up a build script for this is on my todo list. First I have to figure out how the test was originally built though.
2) Interesting, for me the simd version is about ~14s (jsc) / ~16s (v8) on my M4 MBP. The non-simd is ~8/~9s respectively. What CPU are you using? That still might be a bit long so I'm happy to trim the simd iteration count into e.g. half?
3) Let's add that to the doc and talk about it at the next sync.
4) Yeah, I brought this up at the last sync I'm happy to trim the non-SIMD. Although, I think we should keep it until we have enough other wasm tests. At least given the plan to remove the smaller wasm benchmarks already.

danleh:
1. Perfect, I'll take another look with a build script then.
2. I'm running on a server x86 CPU, very high core count but pretty low frequency/boost, that could explain it. Changing iterations to 40 for SIMD SGTM.
3. Sure, I added it to the doc/agenda for next time.
4. Sure, let's keep the non-SIMD for now.

kmiller68:
Ok sounds like we have a plan. Gonna merge for now.

PR #24, opened 2024-12-23
Move argon2 to first, average, worst scoring.

kmiller68:
Since the non-simd version of the test is relatively slow I reduced the iteration/worst counts to 15/2, repsectively.

Also, I moved all the benchmark files to the argon2 directory to help organize the wasm directory a bit.


kmiller68:
CC @danleh @camillobruni @eqrion @jandem for thoughts

danleh:
Regarding the `README.md` (which wasn't changed by this PR, but since I looked at it): Could you add a short description where the source code is coming from? Or ideally change `build.sh` to checkout/download it from some URL, so we could more easily update it with whatever the upstream is?

danleh:
Mhm, I think there is something wrong with the non-SIMD variant of this benchmark. In particular, the flamegraph/profile shows tons of samples attributed to capturing stack traces, caused by growing the memory inside `_emscripten_resize_heap` and `emscripten_realloc_buffer`. See screenshot:
image

If you modify `emscripten_realloc_buffer` in `argon2-bundle.js` to not just swallow the error/failure to grow memory
```
                function emscripten_realloc_buffer(size) {
                    try {
                        wasmMemory.grow(size - buffer.byteLength + 65535 >>> 16);
                        updateGlobalBufferAndViews(wasmMemory.buffer);
                        return 1
                    } catch (e) { console.log(e) }
                }
```
you get tons of errors, e.g., in d8 `RangeError: WebAssembly.Memory.grow(): Maximum memory size exceeded` or in jsc `RangeError: WebAssembly.Memory.grow expects the grown size to be a valid page count`.

I will briefly try to rebuild this from source to diagnose further. @kmiller68 have you seen this before or is this just a problem on my end? Do you have more information on why/how the memory usage of the non-SIMD variant could be higher than the SIMD variant?
For the time being, I don't think the non-SIMD variant is measuring something sensible.

danleh (changes requested):

danleh:
(lone comment, since the file wasn't changed in this PR): I believe the `../` prefix should be stripped from the destination paths in line 17 and 21 of `wasm/argon2/build.sh`, see https://github.com/WebKit/JetStream/blob/ab7d6d00b2eb6ce427aad4a4f3f5ebfb7e594924/wasm/argon2/build.sh#L17

kmiller68:
Closing this in favor of https://github.com/WebKit/JetStream/pull/47

PR #26, opened 2025-01-15
Convert gcc-loops-wasm to JS-style scoring

danleh:
Build with the same options (-O2, Emscripten `-s MODULARIZE`, etc.) as `tsf-wasm` before. Use the `WasmEMCCBenchmark` class introduced earlier. Since we want to run multiple iterations (for first, average, worst scoring), change to small workload size. Each individual iteration still runs >20ms, which we agreed on in the last sync.

Drive-by fixes:
The `WasmEMCCBenchmark` setup code needs to make sure we don't automatically run `main()` after `setupModule()`, hence add `noInitialRun`. The TSF-wasm build script uses `set -o pipefail`, which is a bash option, hence change shebang line. Add Wall time output also for `WasmLegacyBenchmark`.

I compared profiles/flamegraphs before and after the recompilation of `gcc-loops.wasm`. We still spend the vast majority of the CPU samples in top-tier compiled code, which makes sense.


wasm/gcc-loops/gcc-loops.cpp
```diff
@@ -159,10 +159,10 @@ void example9 (unsigned *ret) {
 
 
 /* feature: support data-types of different sizes.
-   Currently only a single vector-size per target is supported; 
-   it can accommodate n elements such that n = vector-size/element-size 
-   (e.g, 4 ints, 8 shorts, or 16 chars for a vector of size 16 bytes). 
-   A combination of data-types of different sizes in the same loop 
+   Currently only a single vector-size per target is supported;
```
danleh, line 162: Whitespace changes, automatically truncated by VS code.

wasm/gcc-loops/gcc-loops.cpp
```diff
@@ -369,18 +369,20 @@ int main(int argc,char* argv[]){
...
+  // BENCH("Example4c",  example4c(N, ia, ib), Mi*2, digest_memory(&ib[0], &ib[N]));
   BENCH("Example7",   example7(4), Mi*4, digest_memory(&a[0], &a[N]));
   BENCH("Example8",   example8(8), Mi/4, digest_memory(&G[0][0], &G[0][N]));
   BENCH("Example9",   example9(&dummy), Mi*2, dummy);
   BENCH("Example10a", example10a(sa,sb,sc,ia,ib,ic), Mi*2, digest_memory(&ia[0], &ia[N]) + digest_memory(&sa[0], &sa[N]));
   BENCH("Example10b", example10b(sa,sb,sc,ia,ib,ic), Mi*4, digest_memory(&ia[0], &ia[N]));
   BENCH("Example11",  example11(), Mi*2, digest_memory(&d[0], &d[N]));
   BENCH("Example12",  example12(), Mi*4, digest_memory(&a[0], &a[N]));
-  //BENCH("Example21",  example21(ia, N), Mi*4, digest_memory(&ia[0], &ia[N]));
+  BENCH("Example21",  example21(ia, N), Mi*4, digest_memory(&ia[0], &ia[N]));
```
danleh, line 385: Note that those sub-items were a bit of a mess before: Even though `Example21` was the only one that was commented out in the source file, I looked at the old binary and realized only a small subset was actually run, namely `Example1`, `Example3`, `Example4a`, `Example7`, `Example10a`, and `Example11`.
@kmiller68 Do you know why this subset was chose in the past?
I think the source file should reflect what we actually run and measure, so I opted for enabling all sub-items (except the four with warnings, see above). Does that make sense to you?

wasm/gcc-loops/benchmark.js
danleh, line 1: Essentially a copy from `wasm/TSF/benchmark.js`.

wasm/gcc-loops/build.sh
danleh, line 1: Based on `wasm/TSF/build.sh`. `_main` is automatically exported by Emscripten, so there is no need to explicitly set `-s EXPORTED_FUNCTIONS=...`.

wasm/gcc-loops/gcc-loops.cpp
```diff
@@ -369,18 +369,20 @@ int main(int argc,char* argv[]){
   BENCH("Example1",   example1(), Mi*10, digest_memory(&a[0], &a[256]));
   BENCH("Example2a",  example2a(N, 2), Mi*4, digest_memory(&b[0], &b[N]));
   BENCH("Example2b",  example2b(N, 2), Mi*2, digest_memory(&a[0], &a[N]));
-  BENCH("Example3",   example3(N, ia, ib), Mi*2, digest_memory(&ia[0], &ia[N]));
-  BENCH("Example4a",  example4a(N, ia, ib), Mi*2, digest_memory(&ia[0], &ia[N]));
-  BENCH("Example4b",  example4b(N-10, ia, ib), Mi*2, digest_memory(&ia[0], &ia[N]));
-  BENCH("Example4c",  example4c(N, ia, ib), Mi*2, digest_memory(&ib[0], &ib[N]));
+  // dlehmann: Disable the following four examples, since Emscripten warns:
```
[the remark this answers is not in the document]
kmiller68: As far as I know wasm doesn't have alignment requirements (outside of SIMD). So it seems fine to leave it, IMO.

wasm/gcc-loops/gcc-loops.cpp
kmiller68: Hmm, I wonder if that was accidental 😬? That seems fine to me though. Although, as stated above I'm happy to leave the above cases since they should be fine within wasm.

danleh:
@kmiller68 and Mozilla folks (e.g., @jandem) any comments or would you be happy to merge this?

JetStreamDriver.js
```diff
@@ -1036,6 +1036,7 @@ class WasmEMCCBenchmark extends AsyncBenchmark {
             let Module = {
                 preRun: [],
                 postRun: [],
+                noInitialRun: true,
```
jandem, line 1039: Is this part of the JS style scoring?

jandem:
LGTM, but CC @eqrion

kmiller68 (approved):
LGTM

kmiller68:
Oops maybe I merged too eagerly. If @eqrion has concerns we can revert or fix separately, sorry.

eqrion:
No concerns here.

PR #27, opened 2025-01-20
Move richards-wasm to JS-style scoring

danleh:
Similar to the line items before:
- Move to its own sub-directory.
- Recompile with an up-to-date Emscripten version, add build script.
- Don't add the benchmarking/timing code into the generated output, but instead keep it in a separate `benchmark.js` file.
- Set an iteration count, such that the overall Wall time isn't too high (before: ~0.7s, now ~2-3s with 20 iterations)
- I checked that the profiles look similar before and after, i.e., we spend the (+-1%) the same proportion of CPU samples in the top tier and the function distribution looks similar.


danleh:
Could you take a look, @kmiller68 @jandem @eqrion? Thanks :)

wasm/richards/benchmark.js
```diff
@@ -0,0 +1,27 @@
...
+    // Instantiate the Wasm module before the first run.
+    if (!Module._setup)
+      await setupModule(Module);
+
+    // Set-up the problem (fill the work queue) on each run.
+    Module._setup();
+
+    // Repeatedly call into Wasm to stress test JS-to-Wasm call performance.
+    // I (dlehmann) suppose this wrapper was added (originally in JetStream 2)
+    // to test that a wrapping JavaScript function doesn't slow things down.
```
jandem, line 16: Mild preference for removing this wrapper function and calling into Wasm directly?

wasm/richards/benchmark.js
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
jandem, line 3: Should this be a JetStream benchmark license header?

jandem:
LGTM with nits addressed or answered.

wasm/richards/benchmark.js
danleh: Sure, I am happy either way, if everyone agrees (I went for maximum consistency with the previous version as a first approach).

wasm/richards/benchmark.js
danleh: Happy to change to a more "standardized" JetStream header. We are not consistent and I don't think we have one. So far, it seems Apple folks use
```
/*
 * Copyright (C) 2024 Apple Inc. All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions
 * are met:
 * 1. Redistributions of source code must retain the above copyright
 *    notice, this list of conditions and the following disclaimer.
 * 2. Redistributions in binary form must reproduce the above copyright
 *    notice, this list of conditions and the following disclaimer in the
 *    documentation and/or other materials provided with the distribution.
 *
 * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
 * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
 * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
 * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
 * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
 * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
 * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
 * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
 * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
 * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
 * THE POSSIBILITY OF SUCH DAMAGE.
*/
```
See e.g. https://github.com/WebKit/JetStream/pull/18/files#diff-daa061b2857d975bb9b022c2a422d74153259d4bbaca4894695605dfc1b046e4,
whereas I used the same as in previous contributions by the V8 team, e.g., in https://github.com/WebKit/JetStream/pull/2/files#diff-ffce3a96d4714be4aee127ca03e82ad410e6eadba9bda818f8c715293c5cef95.

jandem: Thanks, that makes sense. It'd be nice to use the same license headers at some point, but that discussion is unrelated to this PR.

wasm/richards/benchmark.js
kmiller68: Don't recall the context but it was probably added to stress JS->Wasm bridging as that was something we thought / still think is important. I think the assumption is that a reasonable number of applications us the wasm host to drive their workloads. So, I'd prefer we leave it for now and reconsider later.

kmiller68 (approved):
LGTM

wasm/richards/benchmark.js
danleh: Then let's leave it as-is. Jan has voiced only a slight preference, and at least in practical terms, it doesn't seem to matter with Spidermonkey, JSC, or V8 right now (no measurable difference, at least V8 definitely inlines the JS wrapper anyway).

eqrion:
LGTM

danleh:
Thanks for the comments! @kmiller68 looks good to merge :)

PR #28, opened 2025-01-21
Move quicksort-wasm to JS-style scoring

danleh:
The usual changes: use `WasmEMCCBenchmark` class in the runner, rebuild and add build script, use iteration count that makes the Wall time not too high (<2s).
    
In this case, we spend a bit more time in the top-tier than before. However, that line item earlier ran a single <50ms iteration, and had a startup time of ~2-3ms, so I would argue it was very prone to noise, and it's good that we run it for longer.


kmiller68 (approved):
LGTM

danleh:
@jandem @eqrion does that look fine from Mozilla's side?

danleh:
Ah, I just see that @eqrion already gave a :+1: on the original post, so I believe it's good to go.

danleh:
Rebased, @kmiller68 should be good to merge now.

PR #30, opened 2025-01-21
Move sqlite Wasm benchmark to first, worst, average scoring

danleh:
Drive-by: remove unused polyfills, update sqlite source distribution, rebuild.

This now takes about ~6s to run with 15 iterations. I don't want to go much lower with the iteration count, but we could lower `--size` in `argv` further (see `benchmark.js`) to speed up a single iteration. Currently, one iteration takes 300-400ms, so we are well above the threshold we set for ourselves.

@kmiller68 @eqrion @jandem WDYT?


danleh:
Rebased. @kmiller68 If you are fine with it, feel free to merge (Ryan gave his :+1: above).

kmiller68:
Given the iterations are so long I wonder if we should change the workload for this? e.g. to a trace from a real sqlite database. Anyway, we can consider that in the future.

danleh:
If a single iteration is still too long, we can easily change the `size` parameter in `benchmark.js` from 10 to, e.g., 5 or 1. (My assumption is that this upstream benchmark provides more coverage of SQLite features and critical code paths than a single trace, given this benchmark is what the SQLite developers use to optimize the database.)

PR #32, opened 2025-01-22
Add performance.mark/measure to runner

danleh:
...which can help to profile only code which is actually measured, not the setup and driver.

Also some drive-by fixes: Add sqlite3-wasm to `wasm-cli.js` runner, fix typo, use `performance.now` for Wall time of line items.


kmiller68 (approved):
LGTM

PR #33, opened 2025-01-22
Add Dart flute WasmGC workload

danleh:
See README for more details.
One iteration takes about 500-600ms on d8, hence 15 iterations to make it take not too long. Seems to be the first workload that uses ES6 modules (dart2wasm generates one, see `build/flute.dart2wasm.mjs` and `benchmark.js`)

CC @mkustermann


danleh:
Recording offline discussion with @mkustermann, Dart2wasm expert: They expect to land more improvements to the Dart toolchain in the coming weeks/months. So before a major release with this line item included by default, we should make sure to re-build and profile. I also measured locally with d8 that we spend ~10-15% in the GC, which Martin confirmed to be in the ballpark what the native Dart VM does as well.

eqrion:
@danleh @mkustermann Can you confirm if [this issue](https://github.com/dart-lang/flute/issues/23) has been fixed in this benchmark? We ran into it when running this benchmark last year.

mkustermann:
> @danleh @mkustermann Can you confirm if https://github.com/dart-lang/flute/issues/23 has been fixed in this benchmark? We ran into it when running this benchmark last year.

Correct, that was missing. I've just rolled new flute (as well as js shells, binaryen, ...) & recompiled the benchmarks in https://github.com/mkustermann/wasm_gc_benchmarks

danleh:
Thanks Martin! I just updated, so should be fixed now here as well @eqrion. I also briefly double checked the workload: profile/flamegraph looks pretty similar to before, so fine from my side to merge.

eqrion (approved):

danleh:
@kmiller68 Updated to main, passes CI. I also double checked that it runs with a local up-to-date JSC build. Ready to merge from my side.

PR #34, opened 2025-01-22
const variable isn't const

kmiller68:


kmiller68:
Simple fix, just gonna land.

PR #35, opened 2025-01-23
Move HashSet-wasm to first, worst, average scoring

danleh:
Same steps as for the other Wasm line items:
- Move files to subdirectory.
- Rebuild with up-to-date Emscripten.
- Reduce single iteration time, so that we can run multiple iterations (a single iteration before took ~300ms, now ~30ms but repeated 120 times).
- Checked that we spend roughly the same amount of CPU samples in the top-tier and that overall profiles look similar.

Once #28 and this lands, all Wasm line items of JetStream 2 have been moved over to the new scoring. So we can start measuring and (if needed) do a round of adapting workload sizes/iteration counts.

CC @kmiller68 @eqrion


eqrion (approved):

kmiller68 (approved):

kmiller68:
LGTM

PR #36, opened 2025-01-24
Remove Octane-zlib.

kmiller68:
This is another asm.js benchmark that we missed from before. Like float-mm.c, we don't think asm.js has enough in the wild usage to justify benchmarking.


iainireland:
Good catch. LGTM.

kmiller68:
Sorry I messed up and pushed to main instead of the branch. Gotta delete this branch/pr and reopen...

PR #37, opened 2025-01-24
Remove Octane-zlib.

kmiller68:
This is another asm.js benchmark that we missed from before. Like float-mm.c, we don't think asm.js has enough in the wild usage to justify benchmarking.


kmiller68:
https://github.com/WebKit/JetStream/pull/36 was the original PR for this but got messed up...

eqrion (approved):
LGTM

danleh:
Removing asm.js SGTM. However, should we add a Wasm build of that as a replacement, to get some coverage for LZ77-style algorithms?

The CI for Chrome is failing right now, but possibly an infra issue, unrelated to this PR. @camillobruni Could you take a look regarding CI?

camillobruni:
I'll have a look in a moment... likely some accidentally serialized DOM object due to an error is causing issues here.

kmiller68:
Yeah, the CI seems to be broken unfortunately... 

I wouldn't be opposed to some compression algorithm in wasm. I'll add it to our ideas list.

PR #39, opened 2025-01-28
Fix end2end tests

camillobruni:
Don't serialize the JetStreamReady event which broke the chrome driver runs.

### DIFF
--- a/tests/run.mjs
+++ b/tests/run.mjs
@@ -83,6 +83,8 @@ async function testEnd2End() {
     try {
         await driver.get(`http://localhost:${PORT}/index.html?worstCaseCount=2&iterationCount=3`);
         await driver.executeAsyncScript((callback) => {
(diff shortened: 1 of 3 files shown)


tests/run.mjs
```diff
@@ -83,20 +83,25 @@ async function testEnd2End() {
     try {
         await driver.get(`http://localhost:${PORT}/index.html?worstCaseCount=2&iterationCount=3`);
         await driver.executeAsyncScript((callback) => {
-            globalThis.addEventListener("JetStreamReady", callback);
+            globalThis.addEventListener("JetStreamReady", () => callback());
```
kmiller68, line 86: I'm assuming the arrow function somehow fixes the serialization issue? If so it's probably worth a comment so someone doesn't come through later and undo it thinking it's an innocuous refactor.

kmiller68 (approved):
LGTM

PR #40, opened 2025-01-28
Move 8bitbench to first, average, worst scoring

danleh:
- Move all generated files into `build/` subdirectory.
- Use `Benchmark` class in `benchmark.js`, similar to other Wasm line items.
- Reduce number of rendered frames per iteration to keep total line item runtime at around 5s.
- Add build script, rebuild from source.
- Double checked that flamegraph/profile looks the same as before, and same proportion of time is spent in top tier.
- Added function for debugging that dumps a rendered image frame to a PPM file.


danleh:
CC @kmiller68 @eqrion

8bitbench/build/rust/pkg/emu_bench.js
```diff
@@ -1,9 +1,11 @@
+let wasm_bindgen;
+(function() {
+    const __exports = {};
+    let wasm;
```
kmiller68, line 4: Why wrap the body of this script in a function? I assume because this script clobbers the global namespace in a way that breaks another script? Maybe we should have the driver understand the .mjs file hack?

Either way, I'd I think we should keep the indentation consistent. Preferably, we would fix the rest of the file but I'm ok with just not indenting this function.

kmiller68 (commented):
Overall LGTM. I think we should be consistent about indentation though. I'd probably also be worth rebasing so the CI testing works.

8bitbench/build/rust/pkg/emu_bench.js
danleh: That is actually the output of the Rust build step (`wasm-pack build --target no-modules --release` specifically), so an autogenerated file. I would thus not manually modify/re-indent it and hence placed it in the `build/` directory.

danleh:
@kmiller68 Rebased and passes the CI now.

kmiller68 (approved):
LGTM

kmiller68:
@eqrion do you want to give this a look over before merging?

PR #41, opened 2025-01-28
Improve end2end runner log

camillobruni:
Add more test logging to make it easier to debug a failing test.

- Log setup status
- Log incremental browser results

We're currently still lacking basic result validation and checking whether any errors have been thrown.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -279,8 +279,18 @@ class Driver {
             }
 
             benchmark.updateUIAfterRun();
+            console.log(benchmark.name)
 
             if (isInBrowser) {
+                globalThis.dispatchEvent(new CustomEvent("JetStreamBenchmarkDone", {
+                    detail: {
+                        name: benchmark.name,
+                        results: {
+                            score: benchmark.score,
+                            ...benchmark.subScores(),
+                        }
+                    }
+                }));
                 const cache = JetStream.blobDataCache;
                 for (const file of benchmark.plan.files) {
                     const blobData = cache[file];
--- a/tests/run.mjs
+++ b/tests/run.mjs
@@ -1,7 +1,7 @@
 #! /usr/bin/env node
 /* eslint-disable-next-line  no-unused-vars */
 import serve from "./server.mjs";
-import { Builder, Capabilities} from "selenium-webdriver";
+import { Builder, Capabilities } from "selenium-webdriver";
 import commandLineArgs from "command-line-args";
 import commandLineUsage from "command-line-usage";
 
@@ -63,6 +63,7 @@ switch (BROWSER) {
         printHelp(`Invalid browser "${BROWSER}", choices are: "safari", "firefox", "chrome", "edge"`);
     }
 }
+
 process.on("unhandledRejection", (err) => {
     console.error(err);
     process.exit(1);
@@ -76,9 +77,7 @@ const PORT = options.port;
 const server = await serve(PORT);
 
 async function testEnd2End() {
-    const driver = await new Builder()
-        .withCapabilities(capabilities)
-        .build();
+    const driver = await new Builder().withCapabilities(capabilities).build();
     let results;
     try {
         console.log("Preparing JetStream");
@@ -90,7 +89,8 @@ async function testEnd2End() {
             if (globalThis?.JetStream?.isReady)
                 callback()
         });
-        results = await benchmarkResults(driver)
+        results = await benchmarkResults(driver);
+        // FIXME: validate results;
     } finally {
         console.log("\nTests complete!");
         driver.quit();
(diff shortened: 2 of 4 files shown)


tests/run.mjs
```diff
@@ -109,4 +103,39 @@ async function testEnd2End() {
...
+    });
+    await new Promise(resolve => pollIncrementalResults(driver, resolve));
+    const resultString = await driver.executeScript(() => {
+        return JSON.stringify(globalThis.JetStream.resultsObject());
+    });
+    return  JSON.parse(resultString);
+}
+
+const UPDATE_INTERVAL = 250;
+async function pollIncrementalResults(driver, resolve) {
```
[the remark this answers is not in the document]
camillobruni: How did you imagine pushing them from the page to the outer test runner?

I tried to just forward console.log to the outer test runner (this would have been the easiest solution IMO) but somehow that didn't work with selenium (most likely I'm missing something here, but I gave up and resorted to the current approach).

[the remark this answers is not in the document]
kmiller68: I'm not sure how to do it in Selenium. I'll ask internally though.

One option would be to have the JetStreamDriver send `POST` requests to `server.mjs` (like `reportScoreToRunBenchmarkRunner`) and either have the server log those messages or have the runner provide a callback to the server.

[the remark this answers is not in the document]
kmiller68: Unfortunately, it doesn't look like Safari has a way to get the console.log messages. If reading an HTTP POST is annoying you could make a "JetStreamConsoleLog" custom event and do 
```js
const originalLog  = console.log.bind(console);
console.log = function log() {
    globalThis.dispatchEvent(new CustomEvent("JetStreamConsoleLog", {
        detail: arguments.toString();
    }));
    originalLog.call(...arguments);
}
```
In `JetStreamDriver.js`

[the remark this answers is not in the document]
camillobruni: I can give it a shot again with listening to events. The last time I tried you basically don't get any callback within selenium, since the main message loop is fully saturated with jetstream. The main blocking issue was that I couldn't add new globals in Firefox, but I didn't try modifying existing objects.

Probably the custom post-message middleware for LWS is going to be easiest choice here here.

camillobruni:
- Use CSS to mark failed tests red
- Driver now tracks errors
- Add Driver.resultsObjectNew for easier parsing
- Poll resultsObjectNew in the test runner and print only the new tests
- Improve error handling / display in test runner

camillobruni:
Screenshot 2025-03-27 at 15 52 26

JetStreamDriver.js
```diff
@@ -508,7 +528,28 @@ class Driver {
         }
     }
 
-    resultsObject()
+    resultsObject(newStyle = false) {
+        if (newStyle)
+            return this.resultsObjectNew();
+        return this.resultsObjectOld();
```
kmiller68, line 534: Old vs New might not be the best distinction since WebKit's infrastructure relies on the "old" JSON schema and I don't think it's likely to change in the foreseeable future. Can we call it something else? Unfortunately, I don't have a great name... maybe call the old thing `runBenchmarkResultsObject()` (`run-benchmark` is WebKit's benchmark harness script)?

JetStreamDriver.js
```diff
@@ -508,7 +528,28 @@ class Driver {
...
+    }
+
+    resultsObjectNew() {
+        const results = {__proto__: null};
+        for (const benchmark of this.benchmarks) {
+            if (!benchmark.isDone)
+                continue;
```
kmiller68, line 541: IIUC, the only time `isDone` is false is when the benchmark failed right? Maybe we should add `results[benchmark.name] = "FAILED"` to the results object if so? Although, maybe we just abort out when a benchmark fails so this might not even be reachable?

camillobruni: `isDone` is also false for the ones that haven't run yet, this is how I currently get the incremental updates out

- I've added an existing `_state` instance variable to make this more explicit.
- Storting "FAILED" when there are benchmark errors (top-level / setup errors are still captured separately)

JetStreamDriver.js
```diff
@@ -508,7 +528,15 @@ class Driver {
         }
     }
 
-    resultsObject()
+    resultsObject(format = "run-benchmark") {
+        if (format == "run-benchmark")
+            return this.runBenchmarkResultsObject();
+        if (format != "simple")
+            throw Error(`Unknown result format '${format}'`);
+        return this.simpleResultsObject();
```
kmiller68, line 536: Nit: I think a switch would be a bit clearer and easier to extend in the future.

kmiller68 (approved):
LGTM

JetStreamDriver.js
```diff
@@ -438,16 +443,31 @@ class Driver {
...
+    pushError(name, error) {
+        this.errors.push({
+            benchmark: name,
+            error: error.toString(),
+            stack: error.stack
+        });
     }
 
     async initialize() {
+        window.addEventListener("error", (e) => this.pushError("driver startup", e.error));
```
kmiller68, line 470: I think this should be guarded by an `if (isInBrowser)` check.

kmiller68:
Great, LGTM! Thanks!

PR #42, opened 2025-01-29
Add Shell Tests

camillobruni:
Add support for running basic `cli.js` tests with `npm run test:jsc`, `npm run test:spidermonkey`, `npm run test:v8`


camillobruni:
I think this is now working... 
We probably still want to add some additional validation at some point (especially for the web versions of these workloads),

camillobruni:
Going to skip JSC for now, since we get a "rate limit exceeded" warning from querying the amazon buckets... not sure if there is an easier way to install this. I presume we can enable this again after a while and caching will prevent us from breakage.

kmiller68:
Fun fact `jsc` exists on Mac but it's not part of the default `PATH`. You can run it from `/System/Library/Frameworks/JavaScriptCore.framework/Versions/Current/Helpers/jsc`. I think it moved at some point but that was a few years ago so that path should be fine.

camillobruni:
ah cool, ok, I'll add a macos check and use that as the default ther, keeping jsvu as backup for other platforms.

camillobruni:
PTAL again... I've re-added JSC support.
Looks like there might be a wasm failure due to the an older JSC version?

```
  Running Dart-flute-wasm:
  JetStream3 failed: CompileError: WebAssembly.Module doesn't parse at byte 15: 0th type failed to parse because sub types are not enabled
  undefined
 ```

File: tests/run-shell.mjs [hunk 1]
```diff
@@ -0,0 +1,189 @@
...
+          case "win32": return "win";
+          case "darwin": return "mac";
+          case "linux": return "linux";
+          default: throw new Error("Unsupported OS");
+      }
+  };
+  const osArch = () => {
+      switch (os.arch()) {
+          case "x64": return "64";
+          case "ia32": return "32";
```
kmiller68, line 158: Huh, do other JS engines still support x86 32-bit? JSC doesn't anymore at least. Regardless, I assume this isn't actually run since it wouldn't work for JSC.

kmiller68 (approved):
LGTM

tests/run-shell.mjs
camillobruni: v8 definitely does... going to remove that since it might be just more confusing than necessary.

iainireland: SM also supports it, FWIW.

kmiller68:
LGTM

PR #43, opened 2025-01-29
Add zlib-wasm workload

danleh:
As a replacement for the asm.js zlib workload in Octane that we removed.
Based on public domain zpipe.c example from https://www.zlib.net/zlib_how.html.
Compresses a ~60KB Wasm binary, and decompresses it again.
Built with Emscripten, using zlib Emscripten port.
A single iteration takes around 50ms, the total line item around 2s.


danleh:
In https://github.com/WebKit/JetStream/pull/37#issuecomment-2616679889 we discussed adding a compression/LZ77-style replacement workload, now that we removed the asm.js Octane zlib workload. Since there is a zlib port for Emscripten, this is actually quite straightforward, so what do you think of this workload? @kmiller68 @eqrion @jandem

eqrion (approved):
This LGTM. Clever to re-use the wasm binary as the file to compress.

danleh:
@kmiller68 Rebased and ready to be merged from my side.

wasm/zlib/src/main.c
```diff
@@ -0,0 +1,49 @@
...
+
+// To be compiled together with `zpipe.c` from zlib. Forward declarations:
+int def(FILE *source, FILE *dest, int level);
+int inf(FILE *source, FILE *dest);
+void zerr(int ret);
+
+// Exported to JavaScript:
+int compressFile(const char* infilename, const char* outfilename)
+{
+    FILE* infile = fopen(infilename, "rb");
```
kmiller68, line 16: Don't you need to `fclose` these? I assume this will leak memory in the test

kmiller68 (commented):
I think you're missing some `fclose`s, but maybe there's some magic that does it for you? Also, do you know how much time we spend in the `fopen` emscripten glue? It'd be nice to make sure we don't repeat the `async-fs` test where all the time is in `Math.random`.

wasm/zlib/src/main.c
danleh: Oops, indeed. Thanks!

danleh:
> Also, do you know how much time we spend in the fopen emscripten glue? It'd be nice to make sure we don't repeat the async-fs test where all the time is in Math.random.

According to a profile I just took (d8 on x64), we spend ~95.5% of the CPU cycles in Wasm code, ~3% compiling, and <1.5% in JS code, so the Emscripten glue code looks fine to me.

wasm/zlib/benchmark.js
```diff
@@ -0,0 +1,34 @@
...
+      Module._decompressFile(inputzStr, inputzoutStr);
+      if (Module.FS.stat('input').size !== Module.FS.stat('input.z.out').size) {
+        throw new Error("Length after decompression doesn't match");
+      }
+    }
+  }
+
+  validate() {
+    const input = Module.FS.readFile('input');
+    const outputRoundtrip = Module.FS.readFile('input.z.out');
```
kmiller68, line 27: Presumably these don't need to be `free`d since they just return the JS string to you but maybe worth double checking?

wasm/zlib/benchmark.js
```diff
@@ -0,0 +1,34 @@
...
+class Benchmark {
+  async runIteration() {
+    if (!Module._main)
+      await setupModule(Module);
+
+    Module.FS.writeFile('input', Module.wasmBinary);
+
+    const inputStr = Module.stringToNewUTF8('input');
+    const inputzStr = Module.stringToNewUTF8('input.z');
+    const inputzoutStr = Module.stringToNewUTF8('input.z.out');
```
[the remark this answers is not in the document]
danleh: Thanks, done (manual calls to `Module._free`).

wasm/zlib/benchmark.js
danleh: I checked the Emscripten-generated code in `zlib.js` (`FS.readFile` which calls `UTF8ArrayToString`). As you say, it returns a JS string, so nothing to manually free here.

kmiller68 (approved):
LGTM too.

PR #44, opened 2025-01-29
Make Wasm workload sizes more uniform

danleh:
Such that each single iteration takes at least 20ms (to curb undue influence of GC scheduling, noise), but not too long (such that we can run as many iterations as possible in under 2s Wall time). Concretely, this means:
- Reducing work per iteration for tsf-wasm and richards-wasm and sqlite3-wasm, while at the same time increasing number of iterations for those three.
- Reducing HashSet-wasm iterations to 50 (it was the only Wasm line item to run >4s Wall time, and there is already plenty precedent for 50 iterations).
On an x64 workstation, I come out to ~1-1.5s Wall time per Wasm line item. This also has the effect that all scores of Wasm line items are in the same order of magnitude, something around 140-190.

Does this sound reasonable to @kmiller68 and @eqrion & @jandem? If yes, we can then start measuring and comparing the JetStream 2.2 Wasm line items with the old scoring vs. now (`<engine> cli.js -- HashSet-wasm tsf-wasm quicksort-wasm gcc-loops-wasm richards-wasm`).


eqrion (approved):

wasm/richards/richards.c
```diff
@@ -16,9 +16,9 @@
 #include <stdlib.h>
 
 #if 1
-#define                Count           10000*100
-#define                Qpktcountval    2326410
-#define                Holdcountval     930563
+#define                Count           200000
+#define                Qpktcountval    465212
+#define                Holdcountval    186084
```
kmiller68, line 21: I wonder if these should be parameters in the `benchmark.js` file rather than hardcoded? It probably wouldn't make a difference for perf and would make it easier to change.

kmiller68 (approved):
LGTM too.

wasm/richards/richards.c
kmiller68: Let's do this in a follow up if we want to.

danleh: Sorry, didn't get to it this week but can upload a PR with this next one.

PR #45, opened 2025-02-10
Make workload size for richards-wasm a runtime input

danleh:
As requested in https://github.com/WebKit/JetStream/pull/44#discussion_r1945854233

@kmiller68 @eqrion PTAL


kmiller68 (approved):
LGTM, thanks!

PR #51, opened 2025-02-27
Minor cleanups for argon2-wasm

danleh:
Update comment after change, remove unused parameter, remove no longer existing benchmark variant.


kmiller68 (approved):
LGTM, thanks for catching the wasm-cli.js issue and the stale comment. Argument change seems reasonable although I doubt it matters too much.

kmiller68:
Since this un-breaks wasm-cli.js I'm going to merge. We can address any feedback from Mozilla folks in a follow up.

PR #52, opened 2025-02-27
Add the pathname when resolving flute.dart2wasm.mjs

kmiller68:
If the server puts index.html in a subdirectory the current code will fail to resolve so we need to add the pathname.


kmiller68:
This issue breaks our internal A/B infrastructure so I'm just going to merge this. Can address feedback in a follow up. Otherwise we'd have to revert the Dart test but that seems like overkill.

kmiller68:
CC @danleh FYI

PR #53, opened 2025-03-06
Update Dart flute WasmGC workload

mkustermann:
This updates the Dart flute benchmark. Apart from using a newer Dart2Wasm compiler this brings two important changes:

  * The compiler & standard library will now use JavaScript strings as it's only string implementation. It takes advantage of `js-string` builtins if available.

    We anticipate all browsers adopting the `js-string` builtin proposal and optimize the builtin methods.

    NOTE: We currently don't *require* the `js-string` builtin to be available. We polyfill the imported methods and we import string constants from the mjs file. The main reason for this is that JavaScriptCore doesn't support `js-string` builtin yet (filed [0]). Once JSC supports the proposal we want to require it as it allows using the much more efficient string constant import mechanism (where string constants are utf8-encoded in the wasm module instead of importing from JS file)

    [0] https://bugs.webkit.org/show_bug.cgi?id=287402

    (Before the compiler used 3 implementations: `{One,Two}ByteString` and `JSString` and copied the strings across the WasmGC<->JS boundary. But now that we get more performant JS strings we prefer using only JS strings due to avoiding the need to copy them across the boundary, working with JS RegExp, ...)

  * The flute benchmark is now compiled in `-O2` mode that ensures Dart soundness. It will perform runtime type checks as required by the Dart language as well as throw catchable exceptions on out-of-bounds accesses.

    (Before the benchmark was compiled in `-O4` mode which is unsound Dart, it didn't perform the type checks required by the Dart language specification and didn't perform index checks).


danleh:
LGTM

I also briefly measured before and after on d8: It's a tiny bit slower now (I assume because of `-O2` / soundness mode), namely Wall time 9.3s / Total score 7.2 (this PR) vs. Wall time 8.9s / Total score 7.5 (current main). I also took a profile / flame graph, looks roughly the same before and after, no big surprises.

Dart/build.log
```diff
@@ -1,5 +1,5 @@
-Built on 2025-01-28 15:34:10+01:00
+Built on 2025-03-06 09:52:17+01:00
 Cloning into 'wasm_gc_benchmarks'...
-cf32ca4 Recompile all benchmarks
+f80892d Update Dart SDK, switch wasm to -O2, recompile performance benchmarks
```
kmiller68, line 3: Did you manually edit this? This is an output file from `build.sh` normally so next time someone updates they'll see the old log again.

Is there a build with the bits you're uploading here saved somewhere? Maybe we can update that script to pull from there?

kmiller68:
@mkustermann Thanks for the detailed explanation! 

Overall seems reasonable to me (modulo comment on `build.log`). If Dart is planning to default to `-O2` we should probably test that. We plan on implementing `js-string` builtins and given it's polyfillable I imagine folks will use it on the web soon anyway. So the extra 🔥 to get us to implement it sooner rather than later seems reasonable to me.

Dart/build.log
mkustermann: This particular line is the abbreviated commit hash and commit message. See hash & commit message from https://github.com/mkustermann/wasm_gc_benchmarks/commit/f80892dcc9578cb864ebb870f7628c8c825fe890 

That being said I removed one line from the generated `build log`: `git clone` outputed something that's to be interpreted by a terminal rather than normal text (`git clone` shows this `Updating ... (x/y)` where it constantly updates the x/y information of the same line). It wasn't there in the old `build.log` and isn't normal text, so I removed that line

mkustermann:
>  We plan on implementing js-string builtins and given it's polyfillable I imagine folks will use it on the web soon anyway. So the extra 🔥 to get us to implement it sooner rather than later seems reasonable to me.

@kmiller68 As mentioned on the original commit message, it's not entirely polyfill'able because of the magic utf8-encode imports. I'd actually really make this rely on `js-string` builtin being there. If WebKit will get support soon, I can also hold this PR until then. wdyt?

mkustermann:
> `flutter build web` defaults currently to `-O4` (but we intend to switch to `-O2`)

Correction: As I just found out, flutter has already switched to `-O2` mode now (~ 4 weeks ago in  https://github.com/flutter/flutter/commit/30b8eb635bd4e85bf70faccdb2f7a23608173aed)

kmiller68:
> @kmiller68 As mentioned on the original commit message, it's not entirely polyfill'able because of the magic utf8-encode imports. I'd actually really make this rely on js-string builtin being there. 

What's the magic utf8-encode imports? IIUC, js-string builtins is just a list of host provided imports. Is there some other part I'm missing?

> If WebKit will get support soon, I can also hold this PR until then. wdyt?

I don't think we can promise any timeline (per Apple policy). That said, happy to go either way you decide.

mkustermann:
> What's the magic utf8-encode imports? IIUC, js-string builtins is just a list of host provided imports. Is there some other part I'm missing?

Specifically I'm referring to this part of the [specification](https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md#string-constants):

> This proposal adds an extension to the JS-API compile routine to support optimized 'imported string constants'
> to address this use-case.
>
> The `WebAssemblyCompileOptions` dictionary is extended with a `USVString? importedStringConstants` flag.
> ...
> Example
> ```js
> let instance = WebAssembly.instantiate(bytes, {importedStringConstants: "XXX"});
> 
> // The global is automatically populated with the string constant
> assertEq(instance.exports.constant.value, "my string constant");
> ```
> ```wat
>(module
>  (global (import "XXX" "my string constant") (ref extern))
>  (export "constant" (global 0))
> )
> ```

This is a really useful mechanism
* it shrinks the combined wasm+js code size
* it improves startup latency - as less JavaScript code has to be parsed
* it may allow wasm2wasm optimizers that are aware of `js-string` builtin semantics to do a better job optimizing wasm
* ...

eqrion:
@mkustermann Thanks for that explanation, that makes sense to me.

kmiller68:
> Specifically I'm referring to this part of the [specification](https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md#string-constants):

Ah, I missed that part, thanks. I can see why that would be useful.

> The main thing we care about is that this will be included in the next version of JetStream. Is there a timeline when the next version may be published?

I don't think there's a fixed timeline yet. We'll probably finalize the content several (6+) months before publishing. That said, we can't land the polyfill-less version until it works in all browsers. So the choice of whether to land with the polyfill or to wait is kinda up to you.

danleh:
@kmiller68 @eqrion @mkustermann I think we could merge this PR / update this workload now, and the discussion about polyfills/JS string builtins can be taken up again before the "code freeze", right?

eqrion:
@danleh Yeah that works for me.

kmiller68:
SGTM too.

Issue #54, opened 2025-03-11
Dart-flute-wasm doesn't run unless index.html is in the URL


mstange:
This is a regression from #52.

Running https://webkit-jetstream-preview.netlify.app/ stops during Dart-flute-wasm with "Uncaught (in proimse) TypeError: can't access property 1, location.pathname.match(...) is null"

Using https://webkit-jetstream-preview.netlify.app/index.html works.

mstange:
Can you take a look, @kmiller68?

PR #57, opened 2025-03-12
Index.html in the pathname should be optional

kmiller68:
When fixing paths in the url for Dart-flute-wasm I neglected to handle the case where index.html is not in the URL. It should now be robust against any rename of index.html too.

### DIFF
--- a/Dart/benchmark.js
+++ b/Dart/benchmark.js
@@ -269,7 +269,7 @@ class Benchmark {
     if (isInBrowser) {
       // In browsers, relative imports don't work since we are not in a module.
       // (`import.meta.url` is not defined.)
-      let pathname = location.pathname.match(/(.*)index\.html/)[1];
+      const pathname = location.pathname.match(/^(.*\/)(?:[^.]+(?:\.(?:[^\/]+))+)?$/)[1];
       this.dart2wasmJsModule = await import(location.origin + pathname + "./Dart/build/flute.dart2wasm.mjs");
     } else {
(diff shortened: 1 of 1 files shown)


kmiller68:
This seems to have fixed the netify preview issue reported in https://github.com/WebKit/JetStream/issues/54. I tested by loading https://deploy-preview-57--webkit-jetstream-preview.netlify.app/ and running it. @mstange do you want to confirm it works for you too?

Dart/benchmark.js
```diff
@@ -269,7 +269,7 @@ class Benchmark {
     if (isInBrowser) {
       // In browsers, relative imports don't work since we are not in a module.
       // (`import.meta.url` is not defined.)
-      let pathname = location.pathname.match(/(.*)index\.html/)[1];
+      const pathname = location.pathname.match(/^(.*\/)(?:[^.]+(?:\.(?:[^\/]+))+)?$/)[1];
```
mstange, line 272: It might be easier to read if you use `replace`, for example `.replace(/\/index\.html$/, '/')`

mstange:
> @mstange do you want to confirm it works for you too?

I can confirm that https://deploy-preview-57--webkit-jetstream-preview.netlify.app/ completes successfully, thanks!

Dart/benchmark.js
kmiller68: Yeah, although I'd rather do that in a follow up since I'd like to unblock this.

PR #59, opened 2025-03-26
Throw errors in cli.js

camillobruni:
Currently the shells don't fail with an error exit code if the benchmark suite fails.

- Both the d8 and spidermonkey shells fail with a non-zero exit status on unhandled promise rejections.
- JSC currently only prints an error message but exits successfully


kmiller68 (approved):
LGTM

PR #60, opened 2025-03-26
date-format: Add result hashing to prevent DCE

camillobruni:
Dead code elimination could be used to bypass the date formatting in extreme cases.
Create a simple hash from the result strings to prevent this.


kmiller68 (approved):
LGTM

PR #61, opened 2025-03-26
gbemu: Add basic canvas pixel hashing

camillobruni:
Make sure we leak a hash of all "drawn" pixels from the canvas. Currently there isn't a proper side-effect and the whole canvas code could be DCE'd which is rather unrealistic.

### DIFF
--- a/Octane/gbemu-part2.js
+++ b/Octane/gbemu-part2.js
@@ -9783,6 +9783,6 @@ setupGameboy();
 
 class Benchmark {
     runIteration() {
-        runGameboy();
+        this.result = runGameboy();
     }
 }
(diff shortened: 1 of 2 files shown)


Octane/gbemu-part2.js
```diff
@@ -9783,6 +9783,6 @@ setupGameboy();
 
 class Benchmark {
     runIteration() {
-        runGameboy();
+        this.result = runGameboy();
```
kmiller68, line 9786: Nit: Could be worth verifying the result hash matches the expected value.

kmiller68 (approved):
LGTM with nit

camillobruni:
This is also ready for comitting.

PR #62, opened 2025-03-26
Avoid eval in date-format-tofte

camillobruni:
Eval has become less and less popular and is actively prohibited by headers on many domains.
date-format-tofte uses an eval that can be easily replaced with a object lookup.

### DIFF
--- a/SunSpider/date-format-tofte.js
+++ b/SunSpider/date-format-tofte.js
@@ -1,10 +1,3 @@
-function arrayExists(array, x) {
-    for (var i = 0; i < array.length; i++) {
-        if (array[i] == x) return true;
-    }
-    return false;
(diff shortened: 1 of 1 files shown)


iainireland (approved):
LGTM! I was going to bring this eval up in the next meeting; glad to see it disappear.

camillobruni:
There are likely a bunch of other places... but this one was trivial to replace.

iainireland:
The other place where we've seen `eval` as a hotspot is in multi-inspector-code-load, which measures how well you can cache indirect eval. I hate that one less, since an eval cache has non-zero value in the real world.

kmiller68 (approved):
LGTM too.

PR #63, opened 2025-04-01
Jetstream on node js

Shubham9527:
# Running JetStream On Node

## Prerequisites

- Node.js `v22.12.0`
- npm

## Installation

1. Clone the repository
2. Navigate to the repository directory
3. Run `npm run jetstream-node` to start the JetStream on NodeJs

## Things to Note

1. The tests are run in the order they are listed in the `JetStreamDriverWithNode.js` file.
2. All the tests will run at once. currently there is not way to run a single test.

## Output

1. The results are printed in the console for each test like this.
score-screenshot
2. Final result will be printed at the end.
final-result-screenshot


camillobruni:
This approach doesn't scale, copying the complete jetstream runner for node is not a good approach.

If possible can you integrate this into the approach suggested in #42

kmiller68 (changes requested):
Yeah, I don't think this is the right way to do this. I think it would make more sense to have a node shim like we have for the other CLIs.

Shubham9527:
Thanks for your feedback. I'll check #42 PR. will try to implement same approach.

PR #66, opened 2025-04-02
3d-cube-SP: Avoid redeclaring static functions

camillobruni:
A lot of functions in 3d-cube are nested inside `run()` even though they don't close over anything.


kmiller68:
But reinitialization of nested functions is very common in JS because minifiers don't hoist. I would hesitate to remove that from benchmarks unless it's being done in an unrealistic way.

Issue #58, opened 2025-03-13
Consider adding a mode which doesn't use Blobs


mstange:
When running JetStream in the browser, [`prefetchResourcesForBrowser` turns all JS URLs into blob URLs](https://github.com/WebKit/JetStream/blob/901c95b90d74b544e2b8d6bac2a6b86ebe22b43a/JetStreamDriver.js#L833-L842).

These blob URLs make profiling significantly harder.

For example, `https://webkit-jetstream-preview.netlify.app/cdjs/call_sign.js`
might become `blob:https://webkit-jetstream-preview.netlify.app/b151b0cb-69f0-4d3d-a764-80edb3af7d65`

Losing the original path makes it harder to look up the source code based on profiles.
Furthermore, the UUIDs in the Blob URL are different in each run, which makes it harder to compare the runtime of a function between runs.

Could we have a mode which doesn't throw away the JS URLs?

For example, when run with `noPreload` in the query string, i.e. `https://webkit-jetstream-preview.netlify.app/?noPreload`, we could skip the call to `prefetchResourcesForBrowser`.

Constellation:
One of the issue we found before was, if we do some networking during tests, it may stall main process, making CPU p-state lower due to wait for IO, and drastically changing the results. So current Blob approach ensuring downloading all binaries before test running is critical for the correctness of the measurement.

camillobruni:
Yes this now works as expected both in shells and browsers.

PR #70, opened 2025-04-23
Configurable resource prefetching 

camillobruni:
This addresses issue #58 and is useful for local profiling.

If the prefetchResources is disabled, we now no longer preload them eagerly, but rather use the live URL and either script-inject it or use it directly with workers.
This means we can easily get the original source positions for profiling.

- `--no-prefetch` command ilne support
- `&prefetchResources=0` URLParam support
- Improve test coverage and add --no-prefetch tests
- Check that test shell output for failures to catch JSC errors (it does not report exit != 0 on rejected promises)

Drive-by-fix:
- Move FileLoader out of the IIFE
- Check that JetStreamDriver has at least one active benchmark
- Add assertions for valid scores
- Fix benchmarks selection with tags

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
         env:
             GITHUB_ACTIONS_OUTPUT: ""
         strategy:
+            fail-fast: false
             matrix:
                 browser: [chrome, firefox, jsc, safari, spidermonkey, v8]
         steps:


JetStreamDriver.js
```diff
@@ -791,8 +808,12 @@ class Benchmark {
                 addScript(text);
         } else {
             const cache = JetStream.blobDataCache;
-            for (const file of this.plan.files)
-                addScriptWithURL(cache[file].blobURL);
+            for (const file of this.plan.files) {
+                if (globalThis.prefetchResources)
+                    addScriptWithURL(cache[file].blobURL);
+                else
+                    addScriptWithURL(file);
```
kmiller68, line 815: Nit: I would just make this `addScriptWithURL(globalThis.prefetchResources ? cache[file].blobURL : file);`

index.html
```diff
@@ -36,7 +36,9 @@
     const isInBrowser = true;
     const isD8 = false;
     const isSpiderMonkey = false;
+    globalThis.prefetchResources = true;
```
kmiller68, line 39: Ditto on unneeded.

JetStreamDriver.js
```diff
@@ -62,8 +70,12 @@ if (typeof(URLSearchParams) !== "undefined") {
         customTestList = urlParameters.getAll("test");
     globalThis.testIterationCount = getIntParam(urlParameters, "iterationCount");
     globalThis.testWorstCaseCount = getIntParam(urlParameters, "worstCaseCount");
+    globalThis.prefetchResources = getBoolParam(urlParameters, "prefetchResources", true);
```
kmiller68, line 73: I would pass `globalThis.prefetchResources` as the default value rather than true. Seems less likely to cause unexpected behavior if someone manually edits a different part of the script.

JetStreamDriver.js
camillobruni: done

tests/run.mjs
```diff
@@ -60,16 +60,20 @@ const server = await serve(PORT);
 async function runTests() {
     let success = true;
     try {
-        success &&= await runTest("Run Single Suite", () => testEnd2End({ test: "proxy-mobx" }));
-        success &&= await runTest("Run Disabled Suite", () => testEnd2End({ tag: "disabled" }));
-        success &&= await runTest("Run Default Suite", () => testEnd2End());
+        success &&= await runEnd2EndTest("Run Single Suite", { test: "proxy-mobx" });
+        success &&= await runEnd2EndTest("Run Tag No Prefetch",  { tag: "proxy", prefetchResources:"false"});
```
danleh, line 64: just out of interest: any particular reason to run both `proxy` items, not just mbox?

camillobruni: mostly to get some additional coverage.

JetStreamDriver.js
```diff
@@ -280,8 +283,10 @@ class Driver {
         }
 
         const allScores = [];
-        for (const benchmark of this.benchmarks)
-            allScores.push(benchmark.score);
+        for (const benchmark of this.benchmarks) {
+            const score = benchmark.score;
```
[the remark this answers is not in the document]
camillobruni: Added an assertion again here.

PR #71, opened 2025-05-13
Disable all BigInt tests except bigint-noble-ed25519 by default

kmiller68:
We've decided we're only going to keep one BigInt test for now. bigint-noble-ed25519 is the remaining BigInt test enabled by default as it appears to do the most interesting things with BigInts. The rest are now disabled by default. They will run if they're passed via a test list but are otherwise not enabled.

To make this work there's a new property on `Benchmark`s/`plan`s that disables them by default.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -631,6 +631,7 @@ class Benchmark {
         this.tags = new Set(plan.tags);
         this.iterations = getIterationCount(plan);
         this.isAsync = !!plan.isAsync;
+        this.disabledByDefault = !!plan.disabledByDefault;
         this.scripts = null;
         this._resourcesPromise = null;
         this._state = BenchmarkState.READY;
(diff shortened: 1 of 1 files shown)


JetStreamDriver.js
```diff
@@ -2258,25 +2263,28 @@ for (const benchmark of BENCHMARKS) {
...
 {
     const benchmark = benchmarksByName.get(testName);
 
-    if (benchmark)
-        JetStream.addBenchmark(benchmark);
-    else
+    if (!benchmark)
         throw "Couldn't find test named \"" +  testName + "\"";
+
+    if (!forceEnable && benchmark.disabledByDefault)
```
camillobruni, line 2273: nit: would we ever not forceEnable workloads?

camillobruni (approved):
LGTM

iainireland:
LGTM also.

JetStreamDriver.js
kmiller68: We don't but I had the argument to be consistent. I'll remove.

PR #72, opened 2025-05-19
Do not use String object for string functions

Constellation:
In `SunSpider/date-format-xparb.js` and `worker/bomb-subtests/date-format-tofte.js`, we are unnecessarily creating String object from string. But we can just use String. This patch changes `new String(...)` to `String(...)` for stringifying.


kmiller68 (approved):
LGTM

PR #74, opened 2025-05-27
Pass the blob URL for preloads in WasmEMCCBenchmark.

kmiller68:
Right now we pass the original path when loading preloads for WasmEMCCBenchmark. This means we might skip the cache and fetch the content from the network again. We don't want to do this because it means the OS might spin down the CPU, which can punish running faster.

I also moved this logic to the `prerunCode` rather than adding it to the `runnerCode` since logically that makes more sense.

Lastly, have the preloads tuples contain the path for cli runs. This means we no longer have to duplicate the preload paths into the benchmark.js file.

### DIFF
--- a/8bitbench/benchmark.js
+++ b/8bitbench/benchmark.js
@@ -1,4 +1,5 @@
 // Copyright 2025 the V8 project authors. All rights reserved.
+// Copyright 2025 Apple Inc. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
@@ -34,6 +35,16 @@ function dumpFrame(vec) {
 
 class Benchmark {
   isInstantiated = false;
+  romBinary;
+
+  async init() {
+    if (isInBrowser) {
+      let response = await fetch(romBinary);
+      this.romBinary = new Int8Array(await response.arrayBuffer());
+    } else {
+      this.romBinary = new Int8Array(read(romBinary, "binary"));
+    }
+  }
 
   async runIteration() {
     if (!this.isInstantiated) {
@@ -41,7 +52,7 @@ class Benchmark {
       this.isInstantiated = true;
     }
 
-    wasm_bindgen.loadRom(Module.romBinary);
+    wasm_bindgen.loadRom(this.romBinary);
 
     const frameCount = 2 * 60;
     for (let i = 0; i < frameCount; ++i) {
--- a/8bitbench/benchmark.js
+++ b/8bitbench/benchmark.js
@@ -38,12 +38,8 @@ class Benchmark {
   romBinary;
 
   async init() {
-    if (isInBrowser) {
-      let response = await fetch(romBinary);
-      this.romBinary = new Int8Array(await response.arrayBuffer());
-    } else {
-      this.romBinary = new Int8Array(read(romBinary, "binary"));
-    }
+    Module.wasmBinary = await getBinary(wasmBinary);
+    this.romBinary = await getBinary(romBinary);
   }
 
   async runIteration() {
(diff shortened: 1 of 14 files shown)


8bitbench/benchmark.js
```diff
@@ -34,14 +34,24 @@ function dumpFrame(vec) {
 
 class Benchmark {
   isInstantiated = false;
+  romBinary;
+
+  async init() {
```
danleh, line 39: IIUC, this is essentially shared code with the `getBinary` function in `WasmEMCCBenchmark` below, just that there it's only used for initializing the `Module["wasmBinary"] = getBinary(...)` and here it's for an arbitrary (non-Wasm) file, that happens to be the emulator ROM, right?

Can we unify this, ideally also with the mechansim for preloading blogs of JavaScript line items (ARES-6/Babylon below)?

JetStreamDriver.js
```diff
@@ -994,9 +993,16 @@ class Benchmark {
         if (this._resourcesPromise)
             return this._resourcesPromise;
 
-        const filePromises = !isInBrowser ? this.plan.files.map((file) => fileLoader.load(file)) : [];
+        this.preloads = [];
+        this.blobs = [];
 
-        const promise = Promise.all(filePromises).then((texts) => {
+        if (isInBrowser) {
+            this._resourcesPromise = Promise.resolve();
```
danleh, line 1000: I haven't fully understood the (somewhat intertwined, complex) preload, resource loading code. At a high-level, why do we just early return here having added anything to `this.scripts`?

JetStreamDriver.js
```diff
@@ -994,9 +993,16 @@ class Benchmark {
         if (this._resourcesPromise)
             return this._resourcesPromise;
 
-        const filePromises = !isInBrowser ? this.plan.files.map((file) => fileLoader.load(file)) : [];
+        this.preloads = [];
+        this.blobs = [];
```
danleh, line 997: AFAICT, `this.blobs` is never read, can't this be removed?

JetStreamDriver.js
```diff
@@ -1198,58 +1205,42 @@ class WasmEMCCBenchmark extends AsyncBenchmark {
...
-                var xhr = new XMLHttpRequest();
-                xhr.open('GET', path, true);
-                xhr.responseType = 'arraybuffer';
-                xhr.onload = function() {
-                    Module[key] = new Int8Array(xhr.response);
-                    andThen();
-                };
-                xhr.send(null);
+                async function getBinary(key, blobURL) {
+                    const response = await fetch(blobURL);
```
danleh, line 1217: As I said above, this `getBinary` function is only used for the Wasm binary of Emscripten benchmarks, but could we use some more general mechanism from the JavaScript line items?

JetStreamDriver.js
```diff
@@ -1198,58 +1205,42 @@ class WasmEMCCBenchmark extends AsyncBenchmark {
...
             `;
-        }
-
-        str += "}";
 
-        let keys = Object.keys(this.plan.preload);
-        for (let i = 0; i < keys.length; ++i) {
-            str += `loadBlob("${keys[i]}", "${this.plan.preload[keys[i]]}", async () => {\n`;
+        for (let [ preloadKey, blobURLOrPath ] of this.preloads) {
+            if (preloadKey == "wasmBinary") {
```
danleh, line 1232: Can we get rid of this special casing of just the `wasmBinary` preloadKey? There are not that many Emscripten workloads, so I'd be happy to just add a line `Module["wasmBinary"] = genericPreloadFunc(wasmBinary)` to every one of those (for a little less "magic").

PR #76, opened 2025-05-29
ZLib should check for _compressFile rather than _main to see if the module has been instantiated

kmiller68:
Since _main isn't exported by the module we end up recreating the instance/module on each iteration.


kmiller68:
CC @danleh

danleh (approved):

danleh:
@kmiller68 I can't merge this, could you?

kmiller68:
Oops, done.

PR #77, opened 2025-06-05
Use Uint8Array.fromBase64 instead of atob

Constellation:
ECMAScript now has base64 -> Uint8Array conversion function in the standard. Let's just use it instead of relying on DOM's atob.

### DIFF
--- a/wasm/tfjs-model-helpers.js
+++ b/wasm/tfjs-model-helpers.js
@@ -85,13 +85,7 @@ function arrayBufferToBase64(buffer) {
 }
 
 function base64ToArrayBuffer(base64) {
-    var binary_string = atob(base64);
-    var len = binary_string.length;
-    var bytes = new Uint8Array(len);
-    for (var i = 0; i < len; i++) {
-        bytes[i] = binary_string.charCodeAt(i);
-    }
-    return bytes.buffer;
+    return Uint8Array.fromBase64(base64).buffer;
 }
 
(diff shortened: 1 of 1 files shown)


danleh:
As of 2025-07-01 `Uint8Array.fromBase64` is still behind a flag in V8 (`--js-base-64`, see [this test](https://crsrc.org/c/v8/test/mjsunit/harmony/regress/regress-421227542.js) or the [flag definitions here](https://crsrc.org/c/v8/src/flags/flag-definitions.h;drc=c17ee553a8b9c096351810ac43920a22c72c7df9;l=297)).

It is currently staged, so I expect this to become available at some point (before the release of JetStream 3?), but right now from the Chrome/V8 side we are blocked on https://crbug.com/42204568.

danleh:
This should be working now in d8 since the `--js-base-64` flag is now on-by-default. Let's rerun checks and then merge.

danleh:
Ah, too early still. While it landed in d8, not set in Chrome...

anba:
> (Not sure about SpiderMonkey, how's availability there?)

Shipped in Firefox 133. <https://bugzilla.mozilla.org/show_bug.cgi?id=1917885>

camillobruni:
Looks like indeed this is no longer called / used: https://github.com/search?q=repo%3AWebKit%2FJetStream%20base64ToArrayBuffer&type=code

I guess we could just remove it then?

PR #78, opened 2025-07-01
Remove custom run methods

camillobruni:
Given the new tags approach we don't need the custom code to run groups of workloads.


Constellation:
I'll revert it for now as it broke JetStream3 running on the browser.

PR #82, opened 2025-07-03
Use tags for workload selection

camillobruni:
- Remove custom enable-benchmark vars
- Add "default" tag 
- Enable "default" benchmarks

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -404,8 +404,8 @@ class Driver {
         let text = "";
         let newBenchmarks = [];
         for (const benchmark of this.benchmarks) {
-            const id = JSON.stringify(benchmark.constructor.scoreDescription());
-            const description = JSON.parse(id);
+            const description = Array.from(Object.keys(benchmark.subScores())) 
+            description.push("Score")
 
             newBenchmarks.push(benchmark);
             const scoreIds = benchmark.scoreIdentifiers()
@@ -1012,7 +1012,6 @@ class Benchmark {
         return this._resourcesPromise;
     }
 
-    static scoreDescription() { throw new Error("Must be implemented by subclasses."); }
     scoreIdentifiers() { throw new Error("Must be implemented by subclasses"); }
 
     updateUIBeforeRun() {
@@ -1093,10 +1092,6 @@ class DefaultBenchmark extends Benchmark {
             "Worst": this.worst4Score,
             "Average": this.averageScore,
         };
-    }
-
-    static scoreDescription() {
-        return ["First", "Worst", "Average", "Score"];
     }
 
     scoreIdentifiers() {
@@ -1313,10 +1308,6 @@ class WSLBenchmark extends Benchmark {
         };
     }
 
-    static scoreDescription() {
-        return ["Stdlib", "MainRun", "Score"];
-    }
-
     scoreIdentifiers() {
         return ["wsl-stdlib-score", "wsl-tests-score", "wsl-score-score"];
     }
@@ -1489,10 +1480,6 @@ class WasmLegacyBenchmark extends Benchmark {
         };
     }
 
-    static scoreDescription() {
-        return ["Startup", "Runtime", "Score"];
-    }
-
     get startupID() {
         return `wasm-startup-id${this.name}`;
     }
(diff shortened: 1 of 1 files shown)


camillobruni:
gentle ping.

JetStreamDriver.js
```diff
@@ -2357,13 +2352,13 @@ function enableBenchmarksByTag(tag, forceEnable = false)
     }
 
     for (const benchmark of benchmarks) {
-        if (!forceEnable && benchmark.disabledByDefault)
-            continue;
-
+        if (excludeTags && benchmark.hasAnyTag(...excludeTags))
+            continue
         JetStream.addBenchmark(benchmark);
```
kmiller68, line 2357: I think this doesn't deduplicate tests. We should probably push the benchmarks into a `Set` then `Array.from` them at the end. e.g. running

```% jsc cli.js -- default Wasm
Starting JetStream3
Running zlib-wasm:
    Startup: 108.743
    Worst Case: 115.567
    Average: 116.799
    Score: 113.647
    Wall time: 0:01.718
zlib-wasm
Running zlib-wasm:
JetStream3 failed: Error: Cannot run Benchmark zlib-wasm twice
```

This seems like something folks might end up doing relatively easily if/when we add more tags.

kmiller68: Ah, sorry failed to copy that line:
`% jsc cli.js -- default Wasm`

JetStreamDriver.js
```diff
@@ -2371,108 +2374,22 @@ function processTestList(testList)
...
-    runGenerators = false;
-    runSimple = false;
-    runCDJS = false;
-    runWorkerTests = false;
-    runSeaMonster = false;
-    runCodeLoad = false;
-    runWasm = false;
-}
+const defaultDisabledTags = [];
+if (!isInBrowser)
```
kmiller68, line 2386: Nit: I would just check for `globalThis.Worker`. If a CLI adds `Worker` support I assume they'd want these to run.

JetStreamDriver.js
```diff
@@ -2313,6 +2325,8 @@ for (const benchmark of BENCHMARKS) {
         throw new Error(`Duplicate benchmark with name "${name}}"`);
     else
         benchmarksByName.set(name, benchmark);
+    
+    benchmark.tags.add("all");
```
kmiller68, line 2329: This seems unused?

JetStreamDriver.js
```diff
@@ -2371,108 +2374,22 @@ function processTestList(testList)
     else
         benchmarkNames = testList.split(/[\s,]/);
 
-    const forceEnable = true;
     for (const name of benchmarkNames) {
         if (benchmarksByTag.has(name))
```
kmiller68, line 2378: I think you want `name.toLowerCase()`

JetStreamDriver.js
```diff
@@ -2313,6 +2325,8 @@ for (const benchmark of BENCHMARKS) {
         throw new Error(`Duplicate benchmark with name "${name}}"`);
     else
         benchmarksByName.set(name, benchmark);
+    
+    benchmark.tags.add("all");
 
     for (const tag of benchmark.tags) {
```
kmiller68, line 2331: You probably want to `tag = tag.toLowerCase()` this before checking/inserting it into the map.

camillobruni: The conversion is happening in the Benchmark constructor already.
I plan to add some basic unittests for this in a separate PR so we don't accidentally mess this up.

kmiller68 (commented):
LGTM other than the order change

JetStreamDriver.js
```diff
@@ -634,6 +669,14 @@ class Benchmark {
     }
     get isSuccess() { return this._state = BenchmarkState.DONE; }
 
+    hasAnyTag(...tags) {
+        for (const tag of tags) {
+            if (this.tags.has(tag.toLowerCase()))
+                return true;
+        }
+        return false;
```
kmiller68, line 677: Nit: could be `return tags.some((tag) => this.tags.has(tag.toLowerCase()));`

kmiller68 (approved):
LGTM with some nits.

PR #84, opened 2025-07-08
Kotlin Compose Multiplatform WasmGC workload

danleh:
In short, this is a shell-compatible benchmark containing several subitems from the Kotlin/Wasm Compose Multiplatform benchmarks at https://github.com/JetBrains/compose-multiplatform/blob/master/benchmarks/multiplatform/README.md. Compose Multiplatform is a Kotlin UI framework with support for mobile, desktop, and Web. Similar to the Dart/Flutter workload, the actual rendering is stubbed out, and there are two components: the application and layout logic is compiled to WasmGC, whereas low-level drawing is done by [skiko](https://github.com/JetBrains/skiko), a Wasm linear build of Skia. @eymar from JetBrains would be the expert, in case more background and details on the benchmark are needed.

In terms of measurements: On an x64 workstation, running `path/to/d8 cli.js -- Kotlin-compose-wasm` in a ToT d8 takes about 12s with 15 iterations. In the CPU profile, we see ~11% of the cycles spent on Wasm compilation and ~30% in the GC. Around ~7400 functions are compiled lazily with our baseline tier (Liftoff) and ~1400 tiered up to TurboFan, and in the profile the hottest JIT-ed function takes only ~1.5% of the CPU cycles, so I think it's quite a good example of a "large" workload with lots of code exercised. All in all, sounds like it reflects real-world usage (@eymar, please correct me).


danleh:
One issue for the shell version right now is that SpiderMonkey's js shell doesn't implement `setTimeout()`. I didn't want to polyfill that in userspace/JavaScript with `Promise.resolve().then(f)` to not introduce any difference between engines. @iainireland I remember we spoke about this some time ago, would it be possible to implement that?

danleh:
Also right now, this is built with the current [exception handling proposal](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md) using `exnref` and `try_table`, not the [legacy one](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/legacy/Exceptions.md). This seems to be the reason why it's not running in Safari right now (even though it does run with in a local build of jsc shell) @kmiller68.

AFAICT no exceptions are thrown and there is no effect on the CPU profile or binary size either way, so I would be also happy to switch to the legacy EH for now, and recompile later once support for exnref lands in Safari (after the workload freeze, but before a JetStream release).

danleh:
Mhm, I rebuilt after updating the upstream sources and now it works on Safari, maybe I messed up the Gradle build config earlier. In any case, seems like new EH is not a problem :)

eqrion:
I'm working on getting setTimeout into the SM shell [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1976559). I'll probably land it sometime next week. I don't think that's a hard blocker here though, because it'll still work in the Web.

Kotlin-compose/build/compose-benchmarks-benchmarks.uninstantiated.mjs
```diff
@@ -0,0 +1,443 @@
...
+        if (cached !== void 0) return cached;
+        cachedJsObjects.set(ref, ifNotCached);
+        return ifNotCached;
+    }
+
+    const _ref_Li9za2lrby5tanM_ = imports['./skiko.mjs'];
+    const _ref_QGpzLWpvZGEvY29yZQ_ = imports['@js-joda/core'];
+    
+    const js_code = {
+        'kotlin.captureStackTrace' : () => new Error().stack,
```
eqrion, line 17: I'm seeing a significant amount of samples in 'captureStackTrace' here.

The stack seems to have things like:
```
Wasm: Function <compose-benchmarks:benchmarks>.androidx.compose.runtime.LaunchedEffectImpl.onForgotten [baseline] 
Wasm: Function <compose-benchmarks:benchmarks>.androidx.compose.runtime.internal.PlatformOptimizedCancellationException.<init> [baseline] 
Wasm: Function <compose-benchmarks:benchmarks>.kotlin.coroutines.cancellation.CancellationException.<init> [baseline] 
```

Is this expected? It seems unfortunate for UI rendering to require stack trace collection to be fast.

danleh: @eymar Why are we seeing these `CancellationException`s, is this expected?

danleh: If `CancellationException`s are indeed used for "mostly regular" control-flow, that's fine, but does it make sense to materialize a stack trace for them? E.g., is there some option in `build.gradle.kts` that causes this, but is not usually enabled in production builds? Something along those lines? https://github.com/JetBrains/compose-multiplatform/blob/5eaf42c6f9af8fa36575501b50f42f81b32af523/benchmarks/multiplatform/benchmarks/build.gradle.kts#L51 (I tried commenting out that line, but that doesn't change anything wrt. `kotlin.captureStackTrace`, it's is still called >8000 times during the benchmark.)



In other words: Is this a realistic representation of what Kotlin/Wasm programs in the wild are doing? Are there plans to change this in the Kotlin toolchain?

Kotlin-compose/build/compose-benchmarks-benchmarks.uninstantiated.mjs
```diff
@@ -0,0 +1,443 @@
...
+        'androidx.compose.runtime.internal.get_$external_fun' : (_this, p0) => _this.get(p0),
+        'org.jetbrains.skiko.w3c.language_$external_prop_getter' : (_this) => _this.language,
+        'org.jetbrains.skiko.w3c.userAgent_$external_prop_getter' : (_this) => _this.userAgent,
+        'org.jetbrains.skiko.w3c.navigator_$external_prop_getter' : (_this) => _this.navigator,
+        'org.jetbrains.skiko.w3c.performance_$external_prop_getter' : (_this) => _this.performance,
+        'org.jetbrains.skiko.w3c.open_$external_fun' : (_this, p0, p1) => _this.open(p0, p1),
+        'org.jetbrains.skiko.w3c.window_$external_object_getInstance' : () => window,
+        'org.jetbrains.skiko.w3c.now_$external_fun' : (_this, ) => _this.now(),
+        'org.jetbrains.skia.impl.FinalizationRegistry_$external_fun' : (p0) => new FinalizationRegistry(p0),
+        'org.jetbrains.skia.impl.register_$external_fun' : (_this, p0, p1) => _this.register(p0, p1),
```
yurydelendik, line 247: `register` is called with two args, but below `unregister` is used. Per MDN, "if [unregisterToken is] not provided, the target cannot be unregistered."

eymar: Thanks for pointing it out! 
I checked the skiko code. Indeed, `unregister` calls will do nothing.

Fortunatelly, there are no bugs due to this as the clode calls `clean` explicitlly along with `unregister`. Anyway, I added a ticket for us to improve it - https://youtrack.jetbrains.com/issue/SKIKO-1040

eymar: > Though this call is showing up in the profiler as difference in two JS engines

Could you please explain what kind of difference is it?

yurydelendik: > > Though this call is showing up in the profiler as difference in two JS engines
> 
> Could you please explain what kind of difference is it?

I'm still analyzing that. But my reduced micro-benchmark points on use of Wasm CG objects with `register` function and interaction with Minor GCs.

eymar: I see fast 2 solutions right now (requiring no Compose release):

- Edit the function in this PR ```'org.jetbrains.skia.impl.register_$external_fun' : (_this, p0, p1) => _this.register(p0, p1, p0)```

-  Or rebuild the benchmarks from this hotfix branch https://github.com/JetBrains/compose-multiplatform/tree/ok/jetstream3_hotfix 

cc: @danleh

Kotlin-compose/build/compose-benchmarks-benchmarks.uninstantiated.mjs
bashor: Hi everyone! 

Exceptions were changed in the recently released 2.2.20-Beta2. 
TL;DR:
* They read a stack value from a JS Error only when a stack is requested. 
* Each Kotlin Exception stores a reference to a JS Error
   * The JS Error has a back reference to the Kotlin Exception.
   * JS Error's name is modified after creation.
     * In 2.2.20 it will involve small data copying from wasm to js, since we don't use string builtins yet.
     * In 2.3+, it's going to be just writing a reference to JS String.

Kotlin-compose/build/compose-benchmarks-benchmarks.uninstantiated.mjs
danleh: Thanks for providing the hotfix branch, @eymar. I modified `build.sh` to checkout that branch for the time being and rebuilt from source, which then results in `_this.register(p0, p1, p0)`, as desired.

bashor:
> With the updated Kotlin/Wasm toolchain, I am getting some `JsException` errors in jsc shell (and just jsc, not in V8 or SpiderMonkey):
> 
> ```
> $ jsc cli.js -- Kotlin-compose-wasm
> Starting JetStream3
> Running Kotlin-compose-wasm:
> Error: JsException: Exception was thrown while running JavaScript code
> Error: JsException: Exception was thrown while running JavaScript code
> Error: JsException: Exception was thrown while running JavaScript code
> ...
> ```
> 
> @kmiller68 Is there some convenient way / jsc flag to dump a full stack trace whenever an exception is thrown (caught or not)? Or abort on the first exception?
> 
> CC @bashor Possibly the new Kotlin toolchain imports some new JavaScript code that throws (for whatever reason, maybe a missing polyfill?)

@danleh @kmiller68 
I haven't managed to catch all details yet, but it looks like there is some issue with `WebAssembly.JSTag` in jsc. In the new Kotlin toolchain, JSTag is used for all exceptions when it's available at runtime.

kmiller68:
> @kmiller68 Is there some convenient way / jsc flag to dump a full stack trace whenever an exception is thrown (caught or not)? Or abort on the first exception?
Unfortunately, no, not really, I tried looking for uncaught exceptions in the inspector but it appears that they seem to happen repeatedly (both in Safari but also Chrome). This makes it hard to log to anything that even makes it to the inspector, unfortunately. It appears they're thrown from `kotlinx.coroutines.channels.$receiveCOROUTINE$.doResume`.

`JsException` makes me think this is coming from Kotlin itself as `JsException` with that casing doesn't appear in JSC at all and it's in Kotlin's.

> I haven't managed to catch all details yet, but it looks like there is some issue with WebAssembly.JSTag in jsc. In the new Kotlin toolchain, JSTag is used for all exceptions when it's available at runtime.

It's possible there's a bug in our `JSTag` implementation but I wouldn't rule out an issue somewhere else propagating to an uncaught exception inside Kotlin itself. I also tried turning off all our wasm JIT tiers and the issue didn't go away, so at least it doesn't appear to be a JIT bug or anything.

Kotlin-compose/build/compose-benchmarks-benchmarks.uninstantiated.mjs
bashor: @danleh 
>>The current behavior of eager reading a stack value cannot be configured.
>
>What does that mean? We need to update the toolchain (which I did) and the on-demand materialization/reading of the stack trace is going to be the new default?

Yes, it'll be a new default in 2.2.20-Beta2+.

>>In 2.3+, it's going to be just writing a reference to JS String.
>
>When will there be a stable 2.3 Kotlin/Wasm release?

2.2.20 is going to be released in the second half of September.
2.3.0 -- probably by the end of the year.

Constellation:
It may take a bit of time to be available on Safari, so please keep the workaround for now ;) Once we shipped the fix, let you know about that and let's remove it.

danleh:
Let's merge this because I don't think there is anything actionable for us at this point.

After the WebKit fixes are merged and released in Safari, we can drop the hotfix here and update the Kotlin toolchain to stable 2.2.20 (which we should do anyway, once it's out in September).

Does that sound good @kmiller68 @eqrion?

kmiller68:
That sounds reasonable, although I'd like to take another look after uploading the hotfix here.

PR #92, opened 2025-07-11
Add Nutrient benchmark

pweiskircher:
As discussed on our call, a Nutrient benchmark!

This works both in browser and in the shell. These is not the final workload we want to include, just wanted to get a PR up early to make sure this is all looking roughly right.

@danleh Could you give this a look over and tell me if this is approximately what you're looking for? Thank you!


JetStreamDriver.js
```diff
@@ -2248,6 +2248,21 @@ let BENCHMARKS = [
...
+        files: [
+            "./wasm/nutrient/helper.js",
+            "./wasm/nutrient/build/nutrient-viewer.wasm.js",
+            "./wasm/nutrient/benchmark.js",
+        ],
+        preload: {
+            pdfDocument: "./wasm/nutrient/assets/example.pdf",
+            wasmBinary: "./wasm/nutrient/build/nutrient-viewer.wasm",
+        },
+        iterations: 10,
```
danleh, line 2263: I think we can easily increase that to 40 iterations, which gives us potentially less noisy measurement and less incentive to tier-up super aggressively on short running workloads. (On an x64 workstation, even with 40 iterations we are still under 2s Wall time, which is more than fine.)

wasm/nutrient/benchmark.js
```diff
@@ -0,0 +1,212 @@
...
+    setupPrerequisites();
+    injectRequiredGlobals(globalThis);
+  }
+
+  async runIteration() {
+    if (!Module.initPSPDFKit) {
+      await initializeNutrient();
+    }
+
+    Module.FS.writeFile("/document.pdf", await getBinary(pdfDocument));
```
danleh, line 199: If we don't want to measure virtual filesystem IO as part of the workload (probably not, right?), this line should be moved to `init()`.

danleh (commented):
Thanks a lot, that was super quick! Overall it looks great to me. 

I had a quick look at the CPU profile/flamegraph with 40 iterations (see comment), and in d8 we spend ~40% of the cycles compiling. That's quite compilation heavy, so we could consider increasing the runtime of the workload further. We also spend ~20% of the cycles in the top 3 hottest functions, but that LGTM.

CC @kmiller68 @eqrion for input from the other engines.

wasm/nutrient/benchmark.js
danleh: This was now moved into the first iteration, how about moving that into `init()` (which is not measured / part of the score at all)?

pweiskircher: @danleh Correct! We can only write it after initializeNutrient unfortunately.

eqrion:
@danleh I opened up that NPM package and the [license](https://www.npmjs.com/package/@nutrient-sdk/viewer?activeTab=code) points [here](https://cdn.prod.website-files.com/65fdb7696055f07a05048833/670e038ac3a3ef053ea2089d_SDK%20User%20Evaluation%20License%20Agreement.pdf).

I am not a lawyer, but I do not think the terms will be acceptable to be included in this repo.

Specifically:
```
LICENSEE MAY ONLY USE LICENSED TECHNOLOGY FOR EVALUATION PURPOSES, WITH
A VIEW TO PURCHASING A FULL COMMERCIAL DEVELOPMENT LICENSE. IN ALL CASES,
LICENSED TECHNOLOGY'S OBJECT CODE MAY NOT BE SUBMITTED TO APPLE'S APP
STORE, GOOGLE’S PLAY STORE, ANY THIRD PARTY WHATSOEVER, OR MADE
AVAILABLE ON ANY WEBSITE, PRIVATE OR PUBLIC, OR USED IN PRODUCTION. ANY
REDISTRIBUTION OF LICENSED TECHNOLOGY IN EITHER SOURCE OR BINARY FORM IS
STRICTLY PROHIBITED.
3.4. ANY DISTRIBUTION OR MAKING AVAILABLE OF DERIVED WORKS OR LICENSEE
APPLICATIONS TO LICENSEE'S END USERS IS STRICTLY PROHIBITED.
```
```
3.2. Licensee shall not (and is not licensed to):
3.2.1. reverse engineer, decompile, disassemble, bypass any code obfuscation, or otherwise
attempt to derive the source code of any part of the Licensed Technology. (To the
extent that applicable law expressly permits you to decompile the Licensed
Technology in order to obtain information necessary to render the object code libraries
interoperable with other software, you must first obtain written permission from us to
provide the necessary information)
```

I followed up internally at Mozilla and we agreed that we do not want to include any benchmark where we cannot view the source code. It could maybe be okay for a disabled workload that's not part of the main score, but not the main score.

danleh:
Recording offline discussions with Nutrient and in the last JetStream meeting, next steps: Patrik will reach out to their legal team to (potentially) change the license to allow:
- Licensees to use the binary "for benchmarking, performance comparisons of different execution environments, and performance analysis" or something along those lines (not just for evaluating a purchase).
- Hosting a website with the JetStream suite, including this workload, e.g., on https://browserbench.org/JetStream/ (where it could be run manually, e.g., by adding a URL parameter such as ?test=Nutrient-sdk-wasm)
- Putting the binary into the JetStream repository.
- Putting the binary into Chromium, V8, Firefox, SpiderMonkey etc. repositories as part of a vendored copy of the JetStream repo.
- Allow disassembling and analyzing Wasm bytecode for performance investigations.

pweiskircher:
@danleh @eqrion I received a new license from legal that should hopefully cover this usage.

pweiskircher:
> Hi @pweiskircher, are you sure you want to put this in the codebase? You might want to run this past your legal again anyway since other line items in the benchmark are GPL licensed. Apologies that this was missed earlier but I don't want you to be put on a limb.
> 
> I'd also have to run it past our legal too since I'm almost certain the provided license is not GPL compatible. I'm unsure if, that would be sufficient anyway. I think this test would have be removed from the codebase before serving the content.

Thanks for your reply! Yeah, I didn't realize there is GPL benchmarks in this project. Going to check with legal, but you're right, the most likely answer is no.

PR #98, opened 2025-07-29
Add benchmarks argument to Driver

camillobruni:
- Add benchmarks Driver constructor argument
- Move benchmark filtering methods out of the Driver class


JetStreamDriver.js
```diff
@@ -2359,21 +2324,51 @@ function processTestList(testList)
...
 // FIXME: add better support to run Worker tests in shells.
 if (!isInBrowser)
     defaultDisabledTags.push("WorkerTests");
 
 if (typeof testList !== "undefined") {
-    processTestList(testList);
+    benchmarks = processTestList(testList);
 } else if (customTestList.length) {
-    processTestList(customTestList);
+    benchmarks = processTestList(customTestList);
```
danleh, line 2369: possible future nit / clarity question: What's the difference between `testList` and `customTestList`. Could we get rid of the latter, if they basically serve the same purpose?

JetStreamDriver.js
```diff
@@ -2359,21 +2324,51 @@ function processTestList(testList)
     for (let name of benchmarkNames) {
         name = name.toLowerCase();
```
danleh, line 2325: nit: we `toLowerCase()` already inside `enableBenchmarksByTag` and `enableBenchmarkByName` below, so this could be just removed here (or there).

danleh (approved):
Nice cleanup, LGTM with nits.

PR #100, opened 2025-07-29
Merge customTestList and testList

camillobruni:
- Only use globalThis.testList
- Use more consistent globalThis access
- Simplify benchmark params initialisation in shell-config.js


danleh (approved):
LGTM, thanks!

Issue #56, opened 2025-03-11
hello_world-LJF opens a print dialog when run in the browser


mstange:
https://webkit-jetstream-preview.netlify.app/?test=hello_world-LJF opens a print dialog

mstange:
So does list_search-LJF

camillobruni:
Fixed by removing the benchmarks in #104.

Issue #55, opened 2025-03-11
string_lists-LJF doesn't run ("Benchmark is not defined") when run with ?test= syntax


mstange:
Running https://webkit-jetstream-preview.netlify.app/?test=string_lists-LJF fails: "Uncaught ReferenceError: Benchmark is not defined"

mstange:
The same applies to lists-LJF

mstange:
Oh, the LJF tests are not run by default.

PR #104, opened 2025-07-30
Remove LuaJSFight

camillobruni:
These benchmarks haven't been working for a while (see #56 and #55).
Additionally they duplicate the richards workload (already covered in octane) and they are very small and thus highly unrealistic.


danleh (approved):
LGTM, but probably a good idea to get consensus from JavaScript folks, e.g., what does @iainireland say?

iainireland:
I hadn't looked at these tests before, but: oof. Not going to miss them.

LGTM.

PR #111, opened 2025-07-31
Add validate method for async-fs and sync-fs

camillobruni:
Make sure we have global side-effects and validation of the final result.

Drive-by-fix:
- Convert testList items to lowerCase for better ergonomics
- Print error stack if we see remote failures in the end2end tests

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -2394,7 +2394,7 @@ function processTestList(testList)
     for (let name of benchmarkNames) {
         name = name.toLowerCase();
         if (benchmarksByTag.has(name))
-            benchmarks.push(...findBenchmarksByTag(name));
+            benchmarks.concat(findBenchmarksByTag(name));
         else
             benchmarks.push(findBenchmarkByName(name));
     }
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -2391,7 +2391,8 @@ function processTestList(testList)
     else
         benchmarkNames = testList.split(/[\s,]/);
 
-    for (const name of benchmarkNames) {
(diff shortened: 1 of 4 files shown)


JetStreamDriver.js
```diff
@@ -2391,7 +2391,8 @@ function processTestList(testList)
     else
         benchmarkNames = testList.split(/[\s,]/);
 
-    for (const name of benchmarkNames) {
+    for (let name of benchmarkNames) {
+        name = name.toLowerCase();
         if (benchmarksByTag.has(name))
             benchmarks.push(...findBenchmarksByTag(name));
```
kmiller68, line 2397: Nit this could be `benchmarks.concat(findBenchmarksByTag(name))`, which would be slightly more efficient.

generators/sync-file-system.js
```diff
@@ -178,5 +187,16 @@ class Benchmark {
...
         }
+
+        for (let _ of fs.forEachFileRecursively()) {
+            this.totalFileCount++;
+        }
+    }
+
+    validate(iterations) {
+        if (this.totalFileCount != this.EXPECTED_FILE_COUNT * iterations)
+            throw new Error(`Invalid total file count ${this.totalFileCount}`);
```
kmiller68, line 198: Nit: you could include the expected count too.

kmiller68 (approved):
LGTM with nits.

JetStreamDriver.js
camillobruni: done.

generators/sync-file-system.js
camillobruni: done.

PR #113, opened 2025-08-05
Proposal: Group SunSpider into a single workload

camillobruni:
SunSpider has been officially sunset a long time ago, lowering the weight of these workloads seems a realistic request.

instead of just removing it, we can group all the individual workloads into a single benchmark to lower it's priority.


camillobruni:
I still have to deal a bit with the global-state leaks across the workloads. 
Will deal with that tomorow.

PR #114, opened 2025-08-06
Always run all test configurations even if there are failures

camillobruni:
It's easier for development if we see which engines / browser fails and succeds and don't just cancel all test runs if the first one fails.


danleh (approved):
Oh yeah, thanks! I got annoyed by that too...

PR #121, opened 2025-08-11
Update .gitignore

camillobruni:
- Globally exclude node_modules by default
- Ignore v8.log profiling files


danleh (approved):
LGTM (not sure what's up with the Safari failure, because this really cannot cause it...)

PR #125, opened 2025-08-12
Proposal: Limit bomb-worker concurrency

camillobruni:
Provide a more realistic execution pattern work bomb workers.

- start groups of 8 workers in parallel, wait for each worker to complete the startup 
- After starting all workers in groups of 8, execute the worker's workload in groups of 16

- Add dedicated parameters to control concurrency and total number of workers
- Use postMessage to start workloads in each worker


kmiller68:
Is this representative of what sites creating dozens of workers actually do? I would have expected them to do the opposite e.g. create 16 workers but only run 8 at a time.

camillobruni:
Based on what we saw on Photoshop-web (see comments on #106):
- Spawning 8 workers per second up to ~40 in total
- After that there is some work run on some workers (that part I'm not 100% sure anymore what the distribution is, but likely limited on hardware concurrency numbers as well)

In this PR's case, the 16-parallel runs are definitely up for discussion :)

PR #127, opened 2025-08-13
Add --dump-test-list option

Constellation:
This offers an easy way to dump list of tests scheduled for this command invocation. Like,

```
$VM cli.js -- --dump-list-tests
```

Then it dumps list of tests it will execute.


cli.js
```diff
@@ -83,6 +83,7 @@ if ("--help" in cliFlags) {
     console.log("   --iteration-count:   Set the default iteration count.");
     console.log("   --worst-case-count:  Set the default worst-case count");
     console.log("   --dump-json-results: Print summary json to the console.");
+    console.log("   --dump-test-list:    Print test list.");
```
kmiller68, line 86: Nit: might be worth adding "instead of running"

kmiller68 (approved):
LGTM too.

cli.js
Constellation: Sounds good! Changed.

PR #130, opened 2025-08-13
Add Box2D workload compiled with J2CL/J2Wasm to WasmGC

danleh:
Right now, it requires an additional entrypoint in the upstream j2cl benchmark to avoid explicit warmup and perform a constant amount of work (see add-fixed-run-count-api.patch).

Since JSC seems to not implement JS-string builtins yet, I also added a workaround (they never get called in this workload anyway), which we can hopefully remove later. @kmiller68


danleh:
Any comments from Mozilla's side @eqrion and @iainireland?
(Sidenote: I cannot add them as reviewers, unlike you @kmiller68 or Camillo. Could they be missing some permissions / contributor status here?)

kmiller68:
Seems to spend almost half the time in `m_solveVelocityConstraints__void` (~35%) and `m_run__java_lang_Object` (~10%) in both Chrome and Safari. Firefox is a bit more even with ~10% for each but the profiler there seemed to induce a lot of overhead (5x) so idk how much to trust the results.

This isn't necessarily a dealbreaker but I'm not sure this will measure J2Wasm as much as these two functions, which is probably less interesting to benchmark. Happy to land disabled for now while we discuss though.

kmiller68 (approved):
Seems good to me too. Happy to revisit if anything interesting comes up.

PR #133, opened 2025-08-13
Add a new GroupedBenchmark that runs a series of benchmarks.

kmiller68:
Use this for all the SunSpider tests.

Grouped benchmarks just run each sub-benchmark in order then report those back. The score is the geomean of each sub-benchmarks analogous score e.g. the geomean of each sub-benchmark's First or Average scores. All the relevant prefetching is just forwarded to the sub-benchmarks

Only note is that `subScores` re-computes everything every time because it's called before the benchmark runs. Also, tags don't seem to work for the sub benchmarks. This didn't seem important enough to fix right now but we should maybe consider in the future.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -1124,10 +1124,10 @@ class Benchmark {
 class GroupedBenchmark extends Benchmark {
     constructor(plan, benchmarks) {
         super(plan);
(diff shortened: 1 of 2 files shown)


kmiller68:
@iainireland Thoughts?

PR #137, opened 2025-08-14
Only expose a console copy in the realm

camillobruni:
... to avoid accidentally overriding methods on the default console method.
Something that just happened on the new typescript workload #117 .


danleh (approved):
LGTM, thanks!

danleh:
Ah, optional nit: Could you add a comment to the copy line, to explain the rationale?

PR #138, opened 2025-08-14
Kotlin-compose: do not access isD8 global

danleh:
Doesn't change behavior / performance characteristics, since we previously set the global `isD8 = false`, but let's be on the safe side / remove access to that global.

### DIFF
--- a/Kotlin-compose/build.sh
+++ b/Kotlin-compose/build.sh
@@ -4,6 +4,7 @@ set -eo pipefail
 
 # Cleanup old files.
 rm -rf build/
+rm -rf compose-multiplatform/
 
 BUILD_LOG="$(realpath build.log)"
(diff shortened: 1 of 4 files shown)


Kotlin-compose/build.sh
```diff
@@ -4,6 +4,7 @@ set -eo pipefail
 
 # Cleanup old files.
 rm -rf build/
+rm -rf compose-multiplatform/
```
camillobruni, line 7: drive by fix?

camillobruni (approved):

Kotlin-compose/build.sh
danleh: Yes, makes clean rebuilds easier, since you don't need to remove the source checkout manually

PR #143, opened 2025-08-14
New Workload: Babylon class-heavy startup workload

camillobruni:
Applying the same cache-busting pattern, I've created a [Babylon.js](https://github.com/BabylonJS/Babylon.js)-based startup workload.

- Eval a new string on every second iteration (configurable)
- Load all class names from the babylon module
- Set up a basic scene and run a few frames (configurable)

There is an ES6 and ES5 version of the same workload and they both heavily stress parse+top-level setup code with many classes.


camillobruni:
I've now patched setTimeout for both workloads. I think this was accidentally used in the scene workload. Could you double check again whether this is still blocking / adding dead time?

kmiller68 (approved):
LGTM if we get the cli's passing.

iainireland:
LGTM. I need to get some of our tooling updated so that I can do a better job of profiling only the parts of the test in measured time, but from just eyeballing it, it all seems reasonable. I see lots of time spent parsing, anyway, which is what we were going for.

camillobruni:
Should we add signpost support? We've had that internally somewhere to better filter on the time in the measured path?

PR #144, opened 2025-08-14
Convert number 3 to path in logo

### LINKED ISSUE #46: Convert number 3 in SVG logo to path (not text)
The `JetStream3Logo.svg` currently looks differently depending on the OS/browser. E.g., compare Chrome:
Image
vs. Firefox:
Image

This is probably because the number `3` is a `<text>` tag and the referenced font is not available on my machine. Compare that with the name `JetStream` itself, which is a `<path>`.

Could somebody who has the font installed/created the logo fix this? (Probably a 1 minute job in Inkscape, just not very familiar with it.)

danleh:
...to get the same appearance independent of installed fonts. Also run it through SVGO to make it a tiny bit smaller.

Fixes #46.


camillobruni (approved):

kmiller68 (approved):
SGTM

PR #145, opened 2025-08-14
Stop running tfjs benchmarks by default

Constellation:
The desire is retiring these benchmarks, and it is not updated to the newer Wasm benchmark metrics. This patch disables them by default.


kmiller68 (approved):
SGTM

PR #146, opened 2025-08-18
Experimental Workload: Intl

camillobruni:
This could potentially replace the old datetime formatting workloads.
The workload stress tests the following Intl APIs:
- DateTimeFormat
- ListFormat
- RelativeTimeFormat
- NumberFormat
- PluralRules

I've gone with a fairly random choice of locals before spending more time on getting a more representative list.


intl/benchmark.js
```diff
@@ -0,0 +1,344 @@
...
+    } else {
+      const start = performance.now();
+      result = fn();
+      const end = performance.now();
+      console.log(`    ${fn.name}: ${(end - start).toFixed(2)}ms`);
+    }
+    const {lastResult, totalLength} = result;
+    this.lastResult = lastResult;
+    this.totalLength += totalLength;
+  }
```
kmiller68, line 343: Would it make more sense to run this as a GroupedBenchmark?

kmiller68:
Hmm, this seems like it spends about 1/4 to 1/3 of the time in utf8ToBytes, which seems higher than I'd like

camillobruni:
Looks like we get slightly different results perf platform :/. Looking into it.

kmiller68:
Overall LGTM.

> Independent of this, I'm now curious if there are any plans to improve the interop situation here long-term? In my naive understanding I would expect that most JS APIs work consistently across platforms. 

I think TC-39 intentionally left this up to implementations. IIRC, one specific case was for embedded devices where there may not be enough disk space to hold every locale. Also, since most engines/TC-39 don't own ICU the exact localization is implementation defined.

camillobruni:
Ok, then let's definitely keep this disabled. Not sure – should there ever be a fixed agreement on a set of locales we could reconsider this.

PR #147, opened 2025-08-18
New Workload: validatorjs regexp heavy workload

camillobruni:
This runs the [validatorjs](https://github.com/validatorjs/validator.js) test suite and stress-tests various regexp and some dynamic property lookups.

Given that we've lowered the priority of regex-dna by grouping the sunspider workloads, it might make sense to add another regexp workload again.


camillobruni:
Oops, I've profiled before including more tests that skewed the results massively.
I've commented out the IsBase64 test which also removes the need for providing a Buffer fallback in the webpack config.

kmiller68 (commented):
Hmm, I see some samples parsing the benchmark.js with utf-16 source. Can we escape the unicode there? Other options would be to split out the strings into a separate JSON but that seems like more work and makes the test itself harder to read.

PR #149, opened 2025-08-18
New Workload: prismjs source code highlighting

camillobruni:
Startup + syntax highlighting for various source formats with [prismjs](https://github.com/PrismJS/prism)

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -1772,28 +1772,6 @@ let BENCHMARKS = [
         deterministicRandom: true,
         tags: ["Octane", "typescript"],
     }),
-    new AsyncBenchmark({
-        name: "prismjs-startup",
-        files: [
-            "./startup-helper/StartupBenchmark.js",
-            "./prismjs/benchmark.js",
-        ],
-        preload: {
-            // Use non-minified bundle for better local profiling.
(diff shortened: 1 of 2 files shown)


JetStreamDriver.js
```diff
@@ -1818,6 +1818,28 @@ let BENCHMARKS = [
...
+    new AsyncBenchmark({
+        name: "prismjs-startup",
+        files: [
+            "./startup-helper/StartupBenchmark.js",
+            "./prismjs/benchmark.js",
+        ],
+        preload: {
+            // Use non-minified bundle for better local profiling.
+            // BUNDLE: "./prismjs/dist/bundle.es5.js",
+            BUNDLE: "./prismjs/dist/bundle.es5.min.js",
```
kmiller68, line 1830: Is there a reason we're running the es5 over the es6 bundle? Either way, it seems reasonable to have a  non-Default version of the other one.

kmiller68 (approved):
LGTM, with one question

jandem:
LGTM.

JetStreamDriver.js
camillobruni: - added both es5 and es6 variants
- enabled only es6 by default

PR #153, opened 2025-08-19
Update/rewrite web-tooling-benchmark

camillobruni:
The [orignal benchmark](https://github.com/v8/web-tooling-benchmark) has been stale for a while and uses a more node-oriented build process.

Given the experience with the previous webpacked workloads I've decided to simply include the full sources and adapt and rewrite it.


web-tooling-benchmark/README.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+Or you open a web browser and point it to `dist/index.html`, or you can use one
+of the engine JS shells to run the special bundle in `dist/cli.js`. The easiest
+way to install recent versions of the supported JS engine shells is to run
+[`jsvu`](https://github.com/GoogleChromeLabs/jsvu). Afterwards, you can run the
+benchmark as follows:
+
+```sh
+$ chakra dist/cli.js
```
kmiller68, line 84: Does chakra still exist enough to run the benchmark?

web-tooling-benchmark/benchmark.js
```diff
@@ -0,0 +1,34 @@
+
```
kmiller68, line 1: Should probably have a copyright.

kmiller68 (commented):
Overall, LGTM with a comment on the worst case count.

JetStreamDriver.js
```diff
@@ -2564,24 +2564,58 @@ BENCHMARKS.push(new GroupedBenchmark({
...
     "babylon",
     "chai",
-    "coffeescript",
     "espree",
-    "jshint",
-    "lebab",
-    "prepack",
-    "uglify-js",
+    "esprima-next",
+    // "lebab",
```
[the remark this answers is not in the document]
camillobruni: added a comment, and "disabled-by-default" it.

kmiller68 (approved):
LGTM

PR #157, opened 2025-08-20
Add details parameter to conditionally display grouped benchmarks

camillobruni:
This is quite nice for debugging.


camillobruni:
Screenshot 2025-08-20 at 14 00 47

JetStreamDriver.js
```diff
@@ -809,6 +792,26 @@ class Benchmark {
         return code;
     }
 
+    prepareToRun() {
+        const description = Object.keys(this.subScores());
+        description.push("Score");
+
+        const scoreIds = this.scoreIdentifiers();
+        const overallScoreId = scoreIds.pop();
+        let text =  `<div class="benchmark" id="benchmark-${this.name}">
```
danleh, line 801: tiny nit: double space before template string -> format?

JetStreamDriver.js
```diff
@@ -336,28 +339,8 @@ class Driver {
...
-                for (let i = 0; i < scoreIds.length; i++) {
-                    const scoreId = scoreIds[i];
-                    const label = description[i];
-                    text += `<span class="result"><span id="${scoreId}">&nbsp;</span><label>${label}</label></span>`
-                }
-                text += `</p></div>`;
-            }
-        }
+        for (const benchmark of this.benchmarks)
+            text += benchmark.prepareToRun();
```
danleh, line 343: Two nits:
1. `benchmark.prepareToRun` seems to be only relevant for `isInBrowser`, so can we move 342/343 to after the early return?
2. Is there a more meaningful/descriptive name instead of `benchmark.prepareToRun`? `initScoreDiv` or something?

cli.js
```diff
@@ -58,6 +58,8 @@ if ("--ramification" in cliFlags)
     globalThis.RAMification = true;
 if ("--no-prefetch" in cliFlags)
     globalThis.prefetchResources = false;
+if ("--details" in cliFlags)
```
danleh, line 61: Since this is really just about grouped benchmarks, can we give this flag (and the respective variables) a more descriptive name? Something like `--expand-grouped-workloads` or `--group-details` or so?

JetStreamDriver.js
```diff
@@ -1121,7 +1128,7 @@ class Benchmark {
         plotContainer.innerHTML = `<svg width="${width}px" height="${height}px">${circlesSVG}</svg>`;
     }
 
-    updateConsoleAfterRun(scoreEntries) {
+    updateAfterRunInShell(scoreEntries) {
```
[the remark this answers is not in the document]
camillobruni: ups... this was not intended.

JetStreamDriver.js
camillobruni: 1. done
2. changed to `prepareBrowserUI` and moved the benchmarks sorting out into Driver initialize.

JetStreamDriver.js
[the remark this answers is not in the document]
camillobruni: Changed back to updateConsoleAfterRun.

JetStreamDriver.js
```diff
@@ -1062,11 +1062,15 @@ class Benchmark {
 
     updateUIBeforeRun() {
         if (!dumpJSONResults)
-            console.log(`Running ${this.name}:`);
+            this.updateUIBeforeRunInShell();
         if (isInBrowser)
             this.updateUIBeforeRunInBrowser();
     }
 
+    updateUIBeforeRunInShell() {
```
danleh, line 1070: naming nit: This is also not specific to shells, it also prints to the console in the browser. Maybe `updateConsoleBeforeRun`?

JetStreamDriver.js
```diff
@@ -813,6 +793,26 @@ class Benchmark {
         return code;
     }
 
+    prepareBrowserUI() {
```
danleh, line 796: Please excuse another nit: I find it a bit confusing that (1) this by itself doesn't actually prepare or change any UI, since it's stateless and just returning text, and (2) that it has the same name as the `Driver.prepareBrowserUI` which _does_ actually update the UI and has a different signature. Maybe rename this one (and the overridden variant in GroupedBenchmark`) to `getResultDivText` or something (whatever it is, should be different from the Driver method, and make clear that this returns something, not update on its own).

PR #165, opened 2025-08-26
Support custom benchmark arguments

camillobruni:
For some benchmarks its easier to configure certain properties directly from the benchmark config.
See https://github.com/WebKit/JetStream/pull/143 which has different assertion values depending on which sources are used.


danleh (commented):
Two clarifying questions:
1. This is essentially pulling out a small change from #143, right?
2. Is the `Benchmark()` ctor argument not used right now? Because it would go from a plain number (the iteration count) to an object, so not sure if this breaks things.

PR #169, opened 2025-08-29
Status update and CSS fixes

camillobruni:
- Use flexbox and use a uniform var(--gap)
- Only make the workload results section scrollable
- Do not change the item order during running, but rather update the results in place and scroll them into view
- Simplify CSS a bit and avoid most media-size based custom results
- Use uniform min-width between results and start button

https://github.com/user-attachments/assets/2255e43d-87d9-4494-8cfe-5becd5504a70


camillobruni:
Merging this now... happy to follow up and add improvements if this is inconvenient.

PR #171, opened 2025-09-01
Add JetStreamParams

camillobruni:
- Copy params.js from Speedometer
- Implement more uniform params parsing for browser (URLSearchParams) and cli (Map from parsed arguments)
- Acces params/settings via global JetStreamParams
- Expose both JetStreamParams and DefaultJetStreamParams making it east to see if we're running with non-default params


camillobruni:
side-note: once we have this, I can also copy over the dev-menu from speedometer, which makes these parameters a bit more accessible.

PR #173, opened 2025-09-02
Move and share TextEncoder/TextDecoder polyfill

danleh:
Just to reduce some redundancy, and might also be useful for #170 (where another TextDecoder polyfill was about to be added.)


camillobruni:
Ok, a bit nit-picky: instead of `polyfills/fast-text-encoding-1.0.3/text.js` could we use `polyfills/fast-text-encoding/1.0.3/text.js` and put all package meta information in `polyfills/fast-text-encoding`?

danleh:
I'm not sure why jsc/Safari are failing on CI, it does not locally on my x64 Linux machine.

The log at https://github.com/WebKit/JetStream/actions/runs/17401099724/job/49401838880?pr=173#step:8:39 seems to already have an error when jsvu installs and smoke-tests jsc
```
Testing…
  ✖ Error: Command failed with exit code 133 (Unknown system error -133): /Users/runner/.jsvu/bin/javascriptcore /private/var/folders/x7/ch5v91h56_zbvbd1y2f600dm0000gn/T/b478ccbb75b5498d84ea113a99050011
```

@kmiller68 can you reproduce or would know what this could be?

danleh:
JSC seems to be fine again, and Safari CI is failing since multiple weeks, so merging.

PR #175, opened 2025-09-02
More consistent metric reporting

camillobruni:
- Drop the legacy mapping (might be a breaking infra change)?
- Use "Wall-Time" instead of "Wall time"


danleh (approved):
Strong +1 to get rid of this "magic", and _if not now, then when?_ (♫)

PR #176, opened 2025-09-03
Add prettier config

camillobruni:
This PR copies over the prettier config from Speedometer.

Note that I didn't implement the dual prettier + eslint approach, I always found that annoying (we need to duplicate the ignore files) and slow. 

In Speedometer this was introduced to more closely adhere to WebKit's style guide which I always found a bit problematic. I personally introduced a few bugs due to the no-braces for single block statements, so I'd be more than happy to loosen this requirement and go with the prettier defaults for sake of simplicity.

- Single block statements are on the same line
- Braces are required to force a new line for block statements

Beyond that I presonally don't care too much about code formatting, if there is an easy way to get closer to WebKit I'm fine.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,12 @@
+{
+    "arrowParens": "always",
+    "bracketSameLine": false,
+    "bracketSpacing": true,
+    "htmlWhitespaceSensitivity": "css",
+    "printWidth": 250,
+    "semi": true,
+    "singleQuote": false,
+    "tabWidth": 4,
+    "trailingComma": "es5",
+    "useTabs": false
+}
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,7 +3,7 @@
     "bracketSameLine": false,
     "bracketSpacing": true,
     "htmlWhitespaceSensitivity": "css",
-    "printWidth": 250,
+    "printWidth": 100,
     "semi": true,
     "singleQuote": false,
     "tabWidth": 4,
--- a/web-tooling-benchmark/src/acorn.mjs
+++ b/web-tooling-benchmark/src/acorn.mjs
@@ -40,7 +40,11 @@ const payloads = [
   },
 ];
 
-export function runTest(fileData) {
+export function runTest(
+  fileData
+
+
+) {
   const testData = payloads.map(({ name, options }) => ({
     payload: fileData[name],
     name: name,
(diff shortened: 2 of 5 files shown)


web-tooling-benchmark/src/acorn.mjs
```diff
@@ -40,7 +40,11 @@ const payloads = [
   },
 ];
 
-export function runTest(fileData) {
+export function runTest(
```
danleh, line 43: Is this an accidental change?

.prettierrc
```diff
@@ -0,0 +1,12 @@
+{
+    "arrowParens": "always",
+    "bracketSameLine": false,
```
danleh, line 3: This seems to be an HTML/JSX formatting option, do we need/want it in JetStream?

.prettierrc
```diff
@@ -0,0 +1,12 @@
...
+    "bracketSpacing": true,
+    "htmlWhitespaceSensitivity": "css",
```
danleh, line 5: Same with `htmlWhiteSpaceSensitivity`.

danleh (approved):
Generally LGTM with some nits and these high-level questions:

> I didn't implement the dual prettier + eslint approach

Since I'm not familiar with how it's done for Speedometer: Does this mean first running prettier and then (only a subset of) eslint? Keeping it simpler sounds good to me.

> I personally introduced a few bugs due to the no-braces for single block statements, so I'd be more than happy to loosen this requirement and go with the prettier defaults for sake of simplicity.
> - Single block statements are on the same line
> - Braces are required to force a new line for block statements

Strong +1, I also dislike no-braces for single block statements, especially in the combination with blocks with braces, e.g., 
```
if (cond)
  stmt1;
else {
  stmt2;
  stmt3;
}
```

Question 1: Do we plan on having a "big formatting" CL at some point? Or iteratively (per-file or even more fine grained)?

Question 2: Do we plan on having a pre-commit hook in the future? If yes, will it always format everything? (Which would be fine if everything is already in a formatted state, otherwise not so much.)

.prettierrc
camillobruni: we do have the top-level html/css files which we likely want to format too.

kmiller68:
I don't mind braces (were I writing a project from scratch I'd probably expect them). I think JetStream tends to follow WebKit style overall but I personally don't care too much. I'll probably mess up a lot and elide the braces so as long as something complains when I upload the PR I'm fine with whatever.

Let me check if other folks internally care.

PR #178, opened 2025-09-08
Disable superseded Wasm benchmarks by default

danleh:
As discussed in last week's sync. Specifically `gcc-loops` and `quicksort` are very micro-benchmarky, and this reduces the proportion of Wasm workloads again a bit.

(Also moved all disabled/old Wasm workloads to the beginning of the Wasm list now. We can still decide to disable `tsf-wasm` later, since `8bitbench-wasm` is also quite dispatch heavy and possibly a nice replacement.)

@eqrion @kmiller68 Let me know if I can merge this, then we can make some progress on the other open Wasm PRs.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -2093,18 +2093,6 @@ let BENCHMARKS = [
         tags: ["Wasm"],
     }),
     new WasmEMCCBenchmark({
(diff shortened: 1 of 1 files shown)


kmiller68 (approved):
LGTM

PR #182, opened 2025-09-16
Rebuild dotnet workload

### LINKED ISSUE #179: Remove .NET's now dead `BatchSize` variable
I ran into some issues when recompiling the workload after removing `BatchSize`. We should try again with a different/future toolchain. It's possible there's some issue on macOS.

danleh:
and remove obsolete `BatchSize`. Works with 9.0.305 dotnet toolchain on x64 Linux without a problem. Fixes #179.

Runtimes are the same as before on d8 and jsc.


kmiller68 (approved):
LGTM

PR #188, opened 2025-09-19
Dotnet Uses Legacy Wasm Exceptions.

kmiller68:
Add a pass to the build script that replaces them with the non-legacy execptions. This doesn't seem to impact performance. The dotnet runtime still seems to generate one module/function that has a legacy exception but that function is never called. It's also somewhat degenerate since it's just `(try catch_all)` and has no other body.

This doesn't seem to make a difference for the score in all CLIs and we don't want legacy exception usage in JetStream to be the thing that keeps the legacy wasm exceptions "feature" alive.

Note: I didn't run the build script. I just ran the wasm-opt commands on the build products. Mostly because when I recompile the code doesn't run or is 5x slower than the previous binaries.

### DIFF
--- a/wasm/dotnet/build.sh
+++ b/wasm/dotnet/build.sh
@@ -27,6 +27,8 @@ printf '%s\n' 'import.meta.url ??= "";' | cat - ./src/dotnet/bin/Release/net9.0/
 echo "Copying symbol maps..." | tee -a "$BUILD_LOG"
 cp ./src/dotnet/obj/Release/net9.0/wasm/for-publish/dotnet.native.js.symbols ./build-interp/wwwroot/_framework/
 
+# Net9 and Net10 use Emscripten version 3.1.56, which emits legacy EH, see https://github.com/WebKit/JetStream/pull/188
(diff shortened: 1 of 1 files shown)


danleh:
Generally SGTM to avoid depending on legacy EH, but:
1. Let's get some insight from MSFT: @maraf Can we configure dotnet/Blazor to use `exnref` by default? Do you have any plans to change the default? Ideally, I would like to avoid the additional binaryen step (to simplify things and keep them more similar to the real-world build).
2. @kmiller68 Why does it no longer run when you rebuild? Sounds like we should fix the build script for good then.
3. What's the effect of binaryen/`wasm-opt` besides moving to `exnref`? Does it otherwise affect the binaries? (E.g., in terms of code size).

kmiller68:
I looked at all the functions in this list and none of them use exceptions at all (I could have probably saved myself some time and compared that list with the list of functions I posted above but alas...). It's possible one of the functions I listed above is getting inlined here but it seems somewhat unlikely.

Screenshot 2025-11-04 at 10 08 36 AM

PR #189, opened 2025-09-22
Add npm run test:build script

camillobruni:
With the newer workloads we should also check that the build steps keep on working in isolation.


danleh (approved):
LGTM with nits/minor questions.

PR #193, opened 2025-09-30
Fix testIterationCount and testWorstCaseCount params

bvisness:
These params were using the names `iterationCount` and `worstCaseCount`, which are names not used in the code and which therefore result in unexpected behavior in various cases. For example, the cli flags `--iteration-count` and `--worst-case-count` were being totally ignored, and the `DefaultJetStreamParams` fallback behavior in `_parseIntParam` was looking up a nonexistent key.

cc @eqrion


camillobruni:
Closing this since #215 has landed, feel free to reopen if there are further issues.

PR #194, opened 2025-10-02
Remove usage of deprecated std::aligned_storage<>

victorvianna:
This is deprecated in C++23. Use the recommended replacement instead (see crbug.com/388068052).


eqrion:
I think you'll need to re-run the HashSet-wasm build script (build.sh) to get your change to take effect in the benchmark.

victorvianna:
@camillobruni PTAL

kmiller68 (approved):
Seems fine, although this test isn't enabled by default so this may not make much of a difference.

victorvianna:
Could one of you merge the PR?

PR #206, opened 2025-10-16
Improve tags

camillobruni:
- Add stricter tests: tags cannot be benchmark names to avoid accidental selection
- Each benchmark needs to have at least one of "js" or "Wasm" tags
- "Deafult" tag is now lower-case
- Add "bigint" to bigint tests
- Rename "regexp" to "regexp-octane" to avoid clash with the tag of the same name

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -1884,7 +1884,7 @@ let BENCHMARKS = [
         tags: ["default", "js",  "Octane"],
     }),
     new DefaultBenchmark({
-        name: "octane-regexp",
+        name: "regexp-octane",
         files: [
             "./Octane/regexp.js",
         ],


JetStreamDriver.js
```diff
@@ -2793,14 +2793,14 @@ for (const test of INTL_TESTS) {
         iterations: 2,
         worstCaseCount: 1,
         deterministicRandom: true,
-        tags: ["Javascript", "intl"],
+        tags: ["js", "internationalization"],
```
danleh: Do you want to add `...PRISM_JS_TAGS` here as well?

tests/unit-tests.js
```diff
@@ -51,6 +51,23 @@ function assertThrows(message, func) {
...
+    for (const tag of benchmark.tags) {
+      assertFalse(benchmarkNames.has(tag), `'${tag}' is also a benchmark name`);
+    }
+  }
+})();
+
+(function tagsHasWasmOrJS() {
+  for (const benchmark of benchmarks) {
+    const tags = benchmark.tags;
+    assertTrue(tags.has("wasm") || tags.has("js"), `'${benchmark.name}' has no 'js' or 'wasm' tag`);
```
kmiller68: Should this also test for Xor? I guess a test could arguably be both though.

kmiller68 (approved):
LGTM too.

JetStreamDriver.js
```diff
@@ -1858,54 +1858,54 @@ let BENCHMARKS = [
...
         name: "raytrace",
         files: [
             "./Octane/raytrace.js",
         ],
-        tags: ["Default", "Octane"],
+        tags: ["default", "js",  "Octane"],
     }),
     new DefaultBenchmark({
-        name: "regexp",
+        name: "regexp-octane",
```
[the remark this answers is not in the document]
camillobruni: ah yes, SG

tests/unit-tests.js
camillobruni: initially thought about xor this but we definitely want mixed workloads at some point too with wasm-gc.

camillobruni:
There are few more suffixed examples thus stuck with `regexp-octane`.

PR #209, opened 2025-10-17
Await driver.quit

camillobruni:
- Use await "driver.quit()" to properly await the browser and driver shutdown.
- Print safaridriver logs in case of failures (to be removed once we're stable again)

### DIFF
--- a/tests/run-browser.mjs
+++ b/tests/run-browser.mjs
@@ -15,7 +15,6 @@ const optionDefinitions = [
     { name: "help", alias: "h", description: "Print this help text." },
 ];
 
-
 const options = commandLineArgs(optionDefinitions);
 
 if ("help" in options)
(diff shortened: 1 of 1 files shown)


tests/run-browser.mjs
```diff
@@ -163,4 +174,30 @@ function logIncrementalResult(previousResults, benchmarkResults) {
     }
 }
 
+
+function printLogs(sessionId) {
+    if (BROWSER === "safari" && sessionId)
+        return printSafariLogs(sessionId);
+
```
danleh: nit: empty line (also some double newlines above, so reformat)

tests/run-browser.mjs
```diff
@@ -163,4 +174,30 @@ function logIncrementalResult(previousResults, benchmarkResults) {
...
+}
+
+async function printSafariLogs(sessionId) {
```
danleh: Is this logging required for making it work, or is this just for debugging failures in the future?

PR #211, opened 2025-10-21
Add explicit version checks

camillobruni:
- Add utils/version-check.mjs to warn against invalid nodejs versions
- Add license headers to newly added files, utils and test files


danleh (commented):
Regarding the version checks, is there a reason to not use the more builtin checking by adding a `.npmrc` file with `engine-strict=true`? AFAIU node already warns if the versions in `engines` are not fulfilled; with the additional config file it should error out (not tested, just read that at https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json)

Yay to proper license headers :)

PR #212, opened 2025-10-21
Disable transformersjs-whisper-wasm by default.

kmiller68:
Whisper is a much more diverse workload in comparison to Bert but has some notable downsides. In particular, the model is much bigger and its overall memory footprint is much higher. This causes memory issues on iOS and seems to be related to a relatively high jetsam rate there.

Whisper also takes much longer to run than Bert. So getting rid of the longer workload speeds up the benchmark overall.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -2542,7 +2542,7 @@ let BENCHMARKS = [
         iterations: 5,
         worstCaseCount: 1,
         allowUtf16: true,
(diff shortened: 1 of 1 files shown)


kmiller68:
CI / Build error seems unrelated to this PR.

Per sync discussions we're going to deprecate this in favor of BERT.

kmiller68:
> I would say disable instead of deprecate

Ah, yeah, fair enough, disable is definitely more accurate.

PR #217, opened 2025-10-23
sync-fs and async-fs improvements

camillobruni:
- Try to bump min measured iteration duration to ~10ms
- Use generator for randomFileContents
- Only measure setupDirectory in the faster sync-fs
- Increase NOF files and folders
- Add FSVisitor to get some more generator action  going

### DIFF
--- a/generators/async-file-system.js
+++ b/generators/async-file-system.js
@@ -34,9 +34,8 @@ function computeIsLittleEndian() {
 }
 
 const isLittleEndian = computeIsLittleEndian();
-let globalCounter = 0;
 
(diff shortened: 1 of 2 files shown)


kmiller68 (approved):
LGTM, let's get @iainireland to take a look before merging

iainireland:
Looks fine to me. I remain unconvinced that generator performance is particularly important in the real world, but if we're going to benchmark it, we might as well do a good job, and it looks from my quick profiles like this patch succeeds at making the benchmark more generator-heavy.

PR #218, opened 2025-10-23
StartupBenchmark refactoring

camillobruni:
- Move StartupBenchmark and helpers to utils/ (one non-workload folder less at the top-level)
- Migrate web-ssr to use the StartupBenchmark helper


kmiller68 (approved):
Seems reasonable.

PR #223, opened 2025-10-23
Enable jsdom-d3-startup by default.

kmiller68:


kmiller68:
Feel free to just merge if the original PR was wrong and this was meant to be on by default to begin with.

PR #224, opened 2025-10-23
Push up the inner loop counts for the doxbee tests.

kmiller68:
Right now they're quite small and the tests finish in only a few ms. We'd like our tests to run for at least 10ms. After this change the average time in V8, which is the faster browser on this test, runs in about 12/14 ms for async/promise respectively on my M2 Pro Mac.

I also lowered the iteration count to keep the total time for these tests down. From looking at samples they both have a relatively small code set and tier up by the 3-4th iteration. Additionally 80 iterations seems to be long enough to catch the existing worst cases.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -2109,14 +2109,18 @@ let BENCHMARKS = [
         files: [
             "./simple/doxbee-promise.js",
         ],
-        tags: ["default",  "js", "promise", "Simple"],
+        iterations: 80,
+        worstCaseCount: 3,
+        tags: ["default",  "js", "promise", "Simple", "Doxbee"],
     }),
     new AsyncBenchmark({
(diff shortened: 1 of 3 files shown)


iainireland:
IIUC, this increases the total number of inner iterations by 1.5x for doxbee-promise, and 1.67x for doxbee-async? (Outer iterations from 120 -> 80, inner iterations from 10k to 20k/25k). Testing on Firefox and Chrome, it looks like this cuts the reported score roughly in half; not sure why it's not 1.5, but it presumably has to do with the first/worst/average scoring.

We're currently around 10x slower than Chrome (but actively working on it). The increased time is manageable on desktop, but we start to run into slow script dialog problems on slower Android phones in CI. Would it be a problem to reduce the (outer) iteration count further?

PR #225, opened 2025-10-23
Reduce Kotlin-compose-wasm iteration count

kmiller68:
Kotlin-compose-wasm is the longest running test in JS3 by a pretty wide margin. However after the first few iterations almost all code is in the most optimizing compiler in the respective engines.

This change reduces the iterations to 5 and sets adjusts the worst case count accordingly.


kmiller68:
CC @eqrion

kmiller68:
We can discuss offline but those numbers don't seem too bad relative to other `Worst`s in the benchmark. Many of the line items are on the 1000s of samples needed.

kmiller68:
Per discussion at the sync, this seems like an acceptable noise rate for Worst. Will wait until next week to merge unless @cbr comments (or signs off) before then.

PR #235, opened 2025-10-30
More ubuntu tests

camillobruni:
- Run Firefox and Chrome browse tests in headless mode on ubuntu
- Shared common setup steps with yaml anchors to avoid bugs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,7 +92,11 @@ jobs:
         name: Test Build 
         runs-on: ubuntu-latest
         steps:
-            - *checkout-branch
+            - name: Checkout Branch
+              uses: actions/checkout@v4
+              with:
+                fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
+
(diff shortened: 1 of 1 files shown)


danleh (approved):
I suppose that's because there are fewer resource limits on Ubuntu images vs. MacOS, is it? Are we running into these limits or timeouts often (in other words: is this an actual problem)? If we are, LGTM. If we are not, having all three engines run on similar hardware is a somewhat nice rough comparison/smoke test.

danleh:
> We can certainly run jsc CLI tests on Linux. It's essentially identical to our X86_64 Mac implementation, modulo a different ICU CLDR dataset but that shouldn't matter for JetStream.

This we already do (i.e., running jsc on Linux) and I'm certainly fine with. I was more wondering whether it's worth running Firefox and Chrome on Linux (and no longer on MacOS), but as per Camillo's comment

> I just ran into this issue today that we mac jobs were just not scheduled.

it seems this is definitely a good idea as well.

Let's merge then.

PR #236, opened 2025-11-07
Make sure numBytes are integer

Constellation:
It is hard to imagine that using floating-point numbers for numBytes in the wild, which is changed in 6215add877ab. This patch ensures that numBytes are integers.


kmiller68 (approved):
Yeah, seems reasonable to me.

camillobruni:
thanks!

Issue #183, opened 2025-09-16
transformers.js tests are not preloading files, lowering CPU state


Constellation:
It seems that model files and configuration json files are downloaded in the middle of JetStream3 instead of preloading, which can lower CPU state and speed down the browser regardless of JS engine's quality.

<img width="2786" height="508" alt="Image" src="https://github.com/user-attachments/assets/261c2654-ae45-460c-ab52-e6d958575de7" />

danleh:
Oh, we should definitely fix that, thanks for the report! I'll look into it...

danleh:
Unfortunately, I didn't get to finish this and am out of office until mid October. There are two parts to the fix, the first is to provide a custom cache function in the setup during the first iteration of `benchmark.js`, something along those lines:
```
      env.useBrowserCache = false;
      env.useCustomCache = true;
      env.customCache = {
        match: function(url) {
          // Redirect some paths to cached/preloaded resources.
          if (preload[url]) {
            console.log('match preload hit', url, preload[url].byteLength, 'bytes');
            return {
              ok: true,
              status: 200,
              arrayBuffer() { return preload[url]; },
              async blob() {
                return {
                  size: preload[url].byteLength,
                  async arrayBuffer() { return preload[url]; }
                }
              },
            };
          }

          throw new Error('unknown file ' + url + '\n' + 'preload: ' + Object.keys(preload));
        },
        put: function(url, response) {
          return Promise.resolve(undefined);
        },
      }
```

The other issue is that due to the preloading via blob URLs in the driver, the `preload` map will not contain the non-blob-URLs, i.e., this match will fail unless one also adds something like:
```
    preload['./transformersjs/build/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/tokenizer_config.json'] = preload[JetStream.preload['modelTokenizerConfig']];
    preload['./transformersjs/build/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/tokenizer.json'] = preload[JetStream.preload['modelTokenizer']];
    preload['./transformersjs/build/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/config.json'] = preload[JetStream.preload['modelConfig']];
    preload['./transformersjs/build/models/Xenova/distilbert-base-uncased-finetuned-sst-2-english/onnx/model_uint8.onnx'] = preload[JetStream.preload['modelWeights']];
``` 
to `init()`.

[...]

PR #238, opened 2025-11-12
Transformers.js line items shouldn't load resources in the benchmark

kmiller68:
Right now some of the resources are loaded via relative paths in the benchmark. Instead of trying to patch the benchmark to use the blob, this patch makes it easier to redirect `fetch` so that it can find the already loaded data.

I did this by adding a `JetStream.resources` object that maps from the original relative path resource to the blobURLOrPath vended by the preloader.


kmiller68:
This addresses https://github.com/WebKit/JetStream/issues/183

PR #242, opened 2025-12-10
Add BrowserFileLoader

camillobruni:
Similar to ShellFileLoader introduce a BrowserFileLoader as mid-term solution to clean up the file loading code.
- Move common helpers and caches from the driver to the BrowserFileLoader.
- Pass along a "counter" param where easily possible


kmiller68 (approved):
Overall seems fine. I'm a bit mid on the counter passing since it's not fixed everywhere so it's in a weird in-between half global / half parameter state. IMO we should pick one or the other and stick with it or fix it everywhere but I'm not gonna block on it.

PR #245, opened 2025-12-10
Add an Open Governance Policy

kmiller68:
The terms are nearly identical to the ones in Speedometer. This is the same text I shared offline / in the sync.

### DIFF
--- a/Governance.md
+++ b/Governance.md
@@ -2,7 +2,7 @@
 
 JetStream uses multi-stakeholder governance. This will allow us to share work, and build a collaborative understanding of performance on the web in order to drive resourcing towards appropriate areas. This also provides a structure that can endure to provide maintenance and adapt to the future web.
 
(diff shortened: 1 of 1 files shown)


camillobruni:
Minor nit/q: did  you remove the formatting on purpose? Otherwise LGTM from my side (let's wait on daniel too)

camillobruni (approved):

kmiller68:
> Minor nit/q: did you remove the formatting on purpose? Otherwise LGTM from my side (let's wait on daniel too)

Oh... hmm, I'm not sure how it got removed. I'll add it back.

eqrion (approved):
LGTM

danleh (approved):
LGTM with a nit: Shall we change `browser engine with an integrated JavaScript engine` to `browser engine with an integrated JavaScript and WebAssembly engine`, given that the latter is present in more line items now?

(Also sorry for the late reply...)

PR #247, opened 2025-12-11
doxbee-async and doxbee-promise: add side-effects

camillobruni:
Add Queryish.count and use it as simple side-effect in the dummy functions.

### DIFF
--- a/simple/doxbee-promise.js
+++ b/simple/doxbee-promise.js
@@ -116,11 +116,13 @@ module.exports = {
 // a queryish object with all kinds of functions
 function Queryish() {}
 Queryish.count = 0;
-async function dummy_1() {
+function dummy_1() {
   Queryish.count++;
+  return Promise.resolve(undefined);
 }
-async function dummy_2(a) { 
+function dummy_2(a) { 
   Queryish.count++;
+  return Promise.resolve(undefined);
 }
 
 Queryish.prototype.all = dummy_1;


Constellation:
I suggest doing similar in `doxbee-promise` too. Right now, `dummy_1` etc. are not having side-effect much (it is just returning `Promise.resolve(undefined)`.

simple/doxbee-promise.js
```diff
@@ -105,14 +105,24 @@ module.exports = function doxbee(stream, idOrPath) {
...
-
 // a queryish object with all kinds of functions
 function Queryish() {}
+Queryish.count = 0;
+async function dummy_1() {
+  Queryish.count++;
+}
+async function dummy_2(a) { 
+  Queryish.count++;
+}
```
Constellation: They should avoid using `async`, and do
```
function dummy_1() {
  Queryish.count++;
  return Promise.resolve(undefined);
}

function dummy_2(a) { 
  Queryish.count++;
  return Promise.resolve(undefined);
}
```
because that's the purpose of this `doxbee-promise` (not async, that's tested in `doxbee-async`. testing chain of promises).

camillobruni: I'm still observing quite a diff even with this fixed for JSC (almost none for V8):

Before: `321.0 ± 11.0(3.35%)`
After: `309.0 ± 13.0(4.24%)`

PR #251, opened 2026-01-06
Replace prefetch logic

camillobruni:
- Completely asyncify the preload logic
- Move all file loading and caching to BrowserFileLoader
- BrowserFileLoader._blobDataCache now holds temporary promises while the data is still loaded
- BrowserFileLoader._loadBlob now does a retry with timeout in-place without the need for a separate global retry

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -255,11 +255,11 @@ class BrowserFileLoader {
             this._blobDataCache[resource] = blobDataOrPromise;
         }
         const blobData = await blobDataOrPromise;
-        if (globalThis.allIsGood)
-            this._updateCounter();
         // Replace the potential promise in the cache.
         this._blobDataCache[resource] = blobData;
         blobData.refCount++;
+        if (globalThis.allIsGood)
+            this._updateCounter();
         return blobData;
     }
 


JetStreamDriver.js
```diff
@@ -345,14 +336,11 @@ class Driver {
         this.isReady = false;
         this.isDone = false;
         this.errors = [];
+        this.fileLoader = isInBrowser ? browserFileLoader : shellFileLoader;
```
danleh: This seems unused? IIUC, `BrowserFileLoader` and `ShellFileLoader` also have different methods/APIs, so wouldn't work right now anyway. If we only plan on unifying that further in a separate/later PR, I would remove this property and instead add a TODO comment or so.

JetStreamDriver.js
```diff
@@ -208,18 +208,66 @@ class ShellFileLoader {
...
+        return blobURL;
+    }
+    _updateCounter() {
+        ++this.counter.loadedResources;
+        JetStream.updateCounterUI();
     }
 
-    async doLoadBlob(resource) {
-        const blobData = this.blobDataCache[resource];
+    async prefetchResourceFile(file) {
```
danleh: AFAICT this is simply exposing `this._prefetchResource()` under a different name, but there is only one use of `prefetchResourceFile`. Can we simply get rid of this wrapper and rename `BrowserFileLoader._prefetchResource` to `BrowserFileLoader.prefetchResourceFile` (and update the corresponding references)?

JetStreamDriver.js
```diff
@@ -894,6 +884,7 @@ class Benchmark {
 
     get name() { return this.plan.name; }
     get files() { return this.plan.files; }
+    get preloadFiles() { return Object.values(this.plan.preload ?? {}); }
```
danleh: This is probably something for a follow-up cleanup, but I am not a fan of keeping both the plan around and having copies of properties from the plan in the instantiated benchmark as well, e.g., both `this.preloads` and `this.plan.preload` (besides them having slightly different names [singular vs. plural] and types [list vs. object]).

How about using the plan simply as an option bag in the `Benchmark` constructor, initializing the benchmark instance fields with those values, and then never using the plan anymore after that (and also not holding onto it).

JetStreamDriver.js
```diff
@@ -894,6 +884,7 @@ class Benchmark {
```
danleh: (I need to comment here, since GitHub doesn't allow comments on non-changed lines yet. This is referring to the Benchmark constructor just above.)

We should really unify/remove or at least document the difference between:
- Benchmark.plan.files
- Benchmark.plan.preload
- Benchmark.scripts
- Benchmark.preloads
- Benchmark.shellPrefetchedResources
- Benchmark.files
- Benchmark.preloadFiles

Some of them are only available in shell vs. in the browser, some are duplicates, and I certainly don't think all are necessary. Not sure if this is something for this PR though, maybe let's add a TODO with this list in the `Benchmark` constructor?

JetStreamDriver.js
```diff
@@ -255,84 +307,23 @@ class BrowserFileLoader {
...
-
     free(files) {
         for (const file of files) {
-            const blobData = this.blobDataCache[file];
+            const blobData = this._blobDataCache[file];
             // If we didn't prefetch this resource, then no need to free it
             if (!blobData.blob) {
                 continue
             }
             blobData.refCount--;
```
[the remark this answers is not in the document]
camillobruni: oops, added the refCount increment to prefetchResourceFile

PR #252, opened 2026-01-13
Fix dev server output

camillobruni:
- verbose output by default (it's a dev server after all)
- add --quiet option
- Use "undefined" for quite option, instead of "none"  to avoid superfluous "none" outputs

### DIFF
--- a/tests/server.mjs
+++ b/tests/server.mjs
@@ -45,9 +45,7 @@ export const optionDefinitions = [
     },
 ];
 
-export async function serve(options) {
-    let { port, quiet } = options;
-
+export async function serve({ port, quiet }) {
     if (!port) throw new Error("Port is required");
 
     const ws = await LocalWebServer.create({


tests/server.mjs
```diff
@@ -30,27 +30,39 @@ import LocalWebServer from "local-web-server";
...
+        defaultValue: false,
+        description: "Silence the server output.",
+    },
 ];
 
-export async function serve({ port, verbose }) {
-    if (!port)
-        throw new Error("Port is required");
+export async function serve(options) {
+    let { port, quiet } = options;
```
kmiller68: Nit: Why make this change? we don't use options anywhere.

kmiller68 (approved):
> Use "undefined" for quite option, instead of "none" to avoid superfluous "none" outputs
typo: quite -> quiet

LGTM otherwise with nit

PR #254, opened 2026-01-13
The Scatter Plot Should Increase with the Score

kmiller68:
Right now when the score goes up the scatter plots go down. This is a bit confusing since a higher score is better. I think it would be clearer if the plots reflected this.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -1254,7 +1254,7 @@ class Benchmark {
             const result = scores[i];
             const cx = padding + i * xRatio;
             const cy = height - padding - (result - minResult) * yRatio;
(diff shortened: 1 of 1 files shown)


kmiller68:
Screenshot 2026-01-13 at 6 30 06 PM

With the change in this PR the charts look like

camillobruni (approved):

kmiller68:
> I guess the underlying potential confusion here is that the scatter plot shows iteration times not scores

The y-axis is actually derived the scores but inverted. Although practically it's the same thing as the time since they're all normalized. I think having two display units is confusing (this came up when I presented a demo of JS3). IMO, we should stick to one unit throughout the UI. 

FWIW, I'm open to an option to view things as times instead of scores but this is just reflecting the underlying unit correctly.

PR #255, opened 2026-01-14
Improve Benchmark constructor and instance variables

camillobruni:
- Use plan as options bag for constructor arguments
- Prefix "protected" instance variables with "_"
- Rename Benchmark.preload to Benchmark._preloadEntries
- Move iteration and worstCase count processing to the Benchmark class
- Add files and preloadEntries getters so we can override it in GroupedBenchmark
- Rename plan.arguments to plan.args due to "argument" being a reserved word
- Rename WasmLegacyBenchmark to AsyncWasmLegacyBenchmark

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -839,12 +839,11 @@ class Benchmark {
     constructor({
             name, 
             files,
-            preload={},
+            preload = {},
             tags, 
             iterations,
             deterministicRandom = false,
             exposeBrowserTest = false,
-            isAsync=false, 
             allowUtf16 = false,
             args = {} }) {
         this._state = BenchmarkState.READY;
@@ -857,7 +856,6 @@ class Benchmark {
         this.iterations = this._processIterationCount(iterations);
         this._deterministicRandom = deterministicRandom;
         this._exposeBrowserTest = exposeBrowserTest;
-        this.isAsync = !!isAsync;
         this.allowUtf16 = !!allowUtf16;
 
         // Resource handling:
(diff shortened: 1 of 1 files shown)


JetStreamDriver.js
```diff
@@ -855,21 +836,47 @@ class BrowserScripts extends Scripts {
...
-        this.results = [];
+    constructor({
+            name, 
+            files,
+            preload={},
+            tags, 
+            iterations,
+            deterministicRandom = false,
+            exposeBrowserTest = false,
+            isAsync=false,
```
danleh: not quite sure, but I think there is some inconsistency or possible mistake here, since the only mentions of `isAsync` were here in the ctor. Could it be that we used to not have have `AsyncBenchmark` but signaled that via a property instead? The only benchmark plans with a `async` property (but not `isAsync`) are two `WasmLegacyBenchmark`, maybe let's simply remove those two and then the property altogether (in either spelling :D)

JetStreamDriver.js
```diff
@@ -1641,8 +1665,10 @@ class WSLBenchmark extends Benchmark {
 };
 
 class WasmLegacyBenchmark extends Benchmark {
-    constructor(...args) {
-        super(...args);
+    constructor({async= false, ...plan}) {
```
danleh: See above, but I think we should maybe first get rid of WasmLegacyBenchmark altogheter

danleh (approved):
LGTM with nit and questions around `async`/`isAsync`.

danleh (approved):

PR #256, opened 2026-01-14
Update dependencies to fix some npm audit warnings

danleh:
Simply ran `npm audit fix`.

There are 13 remaining because of jsvu, see upstream change https://github.com/GoogleChromeLabs/jsvu/pull/168. Once that lands, we can bump jsvu here as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -107,14 +107,15 @@
             }
         },
         "node_modules/@babel/code-frame": {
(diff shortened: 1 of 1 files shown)


danleh:
Thanks, merging! I'll check for unused dependencies in a follow-up / together with the jsvu bump later.

Issue #86, opened 2025-07-10
Refactor blob preloading code.


kmiller68:
While working on https://github.com/WebKit/JetStream/pull/74 @danleh noted that there seemed to be a bunch of dead code in the preloading code. It does seem like much of the logic there could be simplified.

danleh:
Thanks for filing! Specifically `_loadInternal` can be removed I think.

danleh:
Status update: The dead code is removed with #93. The remaining TODO is to merge/simplify the different code paths for caching resources in the shell vs. browser setting, see
https://github.com/WebKit/JetStream/blob/0b92f932c04062c597e1690d21ad87b2c8ef9aa1/JetStreamDriver.js#L192-L195
vs
https://github.com/WebKit/JetStream/blob/0b92f932c04062c597e1690d21ad87b2c8ef9aa1/JetStreamDriver.js#L496-L500

@kmiller68: Before refactoring this further, I'd like to understand the existing code a little better. E.g., why is there a retry mechanism in `doLoadBlob`? This feels like something that should not be necessary with a reliable webserver?
https://github.com/WebKit/JetStream/blob/0b92f932c04062c597e1690d21ad87b2c8ef9aa1/JetStreamDriver.js#L840-L861

kmiller68:
@kmiller68: Before refactoring this further, I'd like to understand the existing code a little better. E.g., why is there a retry mechanism in doLoadBlob? This feels like something that should not be necessary with a reliable webserver?

I believe this was added because folks complained the benchmark didn't run on spotty Wifi/cellular connections.

PR #259, opened 2026-01-14
Merge Browser and Shell file and script loader paths

camillobruni:
- Use common interface methods for BrowserFileLoader and ShellFileLoader
- Add missing globalThis.allIsGood  for shell config
- Move file access snippets to BrowserScript and ShellScript

Issue: #86

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -279,12 +279,13 @@ class BrowserFileLoader extends FileLoader {
     }
 
     async _fetchBlobData(blobData) {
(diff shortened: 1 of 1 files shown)


JetStreamDriver.js
```diff
@@ -806,6 +868,24 @@ class BrowserScripts extends Scripts {
         this.add("window.onerror = top.currentReject;");
     }
 
+    static get prerunCode() {
+        return `
+            JetStream.getBinary = async function(blobURL) {
```
camillobruni: verbatim copy

PR #262, opened 2026-01-17
Replace prefetch logic (#251) is a 2% regression

kmiller68:
A/B testing indicates this change caused a 2% score change in Safari. This seems surprising since this was intended to be a purely harness change. Reverting for now to confirm and to investigate the cause.

This also reverts "Add resource loading bar (#246)" and "Improve Benchmark constructor and instance variables (#255)" as they blocked a clean revert.

For posterity, this is tracked internally as rdar://problem/168103740

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -56,6 +56,25 @@ function displayCategoryScores() {
     document.body.classList.add("details");
 }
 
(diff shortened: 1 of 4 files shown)


danleh:
@kmiller68 Can you reproduce this regression locally, ideally even on JSC/shells?

PR #265, opened 2026-01-19
Reland: Replace prefetch logic

camillobruni:
- Completely asyncify the preload logic
- Move all file loading and caching to BrowserFileLoader
- BrowserFileLoader._blobDataCache now holds temporary promises while the data is still loaded
- BrowserFileLoader._loadBlob now does a retry with timeout in-place without the need for a separate global retry

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -307,8 +307,9 @@ class BrowserFileLoader {
     }
 }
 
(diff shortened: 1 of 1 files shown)


camillobruni:
Given that we've now forked the v3.0 branch we should be good to land this again?

PR #266, opened 2026-01-20
Fix --dump-test-list

camillobruni:
We moved the params parsing partially to the Params which was not use in cli.js
Directly use the cli-only arguments in cli.js instead.

### DIFF
--- a/tests/helper.mjs
+++ b/tests/helper.mjs
@@ -37,6 +37,7 @@ export function logInfo(...args) {
 }
 
 export function logError(...args) {
+  let error;
   if (args.length == 1 && args[0] instanceof Error)
     error = args[0];
   const text = args.join(" ");


cli.js
```diff
@@ -109,12 +109,15 @@ function parseCliFlag(argument) {
         throw Error(message);
     }
     let value = parts.slice(1).join("=");
-    if (flagName === "no-prefetch") value = "false";
+    if (flagName.startsWith("no-")) value = "false";
```
danleh: Do we have some other parameter that uses this `--no` prefix? Or rather: all other boolean parameters are set to false by default, whereas `prefetchResources` is true, right? So maybe don't generify this just yet, since supplying, e.g., `--no-ramification` does nothing, right?

PR #268, opened 2026-01-22
dotnet: move the Wasm compilation into the first run

danleh:
As discussed yesterday, I think we should have all "interesting" or substantial parts of the workload inside the measurement window. So far, the dotnet Wasm compilation (as part of `this.dotnet.withModuleConfig`) was not, since it happened in `Benchmark.init()`.

This PR moves it into the first iteration, as it should have been (and is for the other Wasm workloads).

CC @eqrion @kmiller68

### DIFF
--- a/wasm/dotnet/benchmark.js
+++ b/wasm/dotnet/benchmark.js
@@ -1,6 +1,6 @@
 class Benchmark {
     async init() {
-        const config = {
+        this.config = {
             mainAssemblyName: "dotnet.dll",
             globalizationMode: "custom",
             assets: [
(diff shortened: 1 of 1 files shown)


kmiller68 (approved):
LGTM and I got the OK to merge this onto the branch. Once @eqrion signs off we can land/cherry-pick

eqrion (approved):
LGTM, I agree with this change. Thanks!

danleh:
Thanks, perfect. I'll merge this to main in the UI and upload a cherry-pick to 3.0 in a bit.

PR #274, opened 2026-02-05
Prepare in-depth.html work

camillobruni:
- Format in-depth.html
- Add empty in-depth.html sections for the new workloads
- Add unit tests to check that all benchmarks have a matching in-depth entry

First steps to address #199.

Issue #199, opened 2025-10-13
Update subtest descriptions / help texts


camillobruni:
Many entries in in-depth.html are outdated or need to be added for new workloads.

PR #275, opened 2026-02-06
Add Wasm descriptions to in-depth.html, grey out non-default workloads

danleh:
This is based on top of https://github.com/WebKit/JetStream/pull/274, so let's merge that one first. @camillobruni 

This PR:
- Adds descriptions for all Wasm workloads.
- Extends the test to check that the `<dt>` tag text contents matches the `id="..."` id property.
- Adds a CSS style for non-default workloads (slightly grayed-out).
- Extends the text to check that all non-default workloads from the `JetStreamDriver.js` have this CSS class set.
- (very minor drive-by:) Add some padding at the bottom of in-depth.html

Fixes the wasm part of https://github.com/WebKit/JetStream/issues/199

### DIFF
--- a/in-depth.html
+++ b/in-depth.html
@@ -88,7 +88,7 @@
             </p>
 
             <p>
-                All but one of JetStream 3's benchmarks run for N iterations, where N is often 120.
+                All but two of JetStream 3's benchmarks run for N iterations, where N is often 120.
                 JetStream 3 calculates the startup score from the time it takes to run the first iteration.
                 The worst case score is the average of the worst M iterations, excluding the first iteration.
                 M is always less than N, and is usually 4.
@@ -142,7 +142,7 @@
             </h3>
 
             <p>
-                The greyed-out workloads are not run by default but can be manually enabled on the command-line or via
+                The <span style="color: var(--text-color-very-subtle);">greyed-out workloads</span> are not run by default but can be manually enabled on the command-line or via
                 the <i>testList</i> URL parameter.
             </p>
 


tests/run-browser.mjs
```diff
@@ -238,6 +254,67 @@ async function benchmarkResults(driver) {
     return JSON.parse(resultsJSON);
 }
 
+async function inDepthPageTest(driver) {
+    await driver.get( `http://localhost:${PORT}/in-depth.html`);
```
kmiller68:
```suggestion
    await driver.get(`http://localhost:${PORT}/in-depth.html`);
```

in-depth.html
```diff
@@ -32,769 +33,860 @@
...
+                For most of the JavaScript and WebAssembly benchmarks in JetStream 3, individual scores equally weigh
+                startup performance, worst case performance, and average case performance.
+                These three metrics are crucial to running performant JavaScript and WebAssembly in the browser.
+                Fast startup times lead browsers to loading pages more quickly.
+                Good worst case performance ensures web applications can run without hiccups.
+                Fast average case performance makes it so that the most advanced web applications can run at all.
+            </p>
+
+            <p>
+                All but one of JetStream 3's benchmarks run for N iterations, where N is often 120.
```
kmiller68: This seems incorrect now since typescript (any others?) is also a run once benchmark.

kmiller68 (approved):
LGTM with some suggestions

in-depth.html
danleh: Ah, I didn't know of any others. Typescript also seems to be the only one with `worstCaseCount: 0`, so I changed this to `All but two`.

PR #279, opened 2026-02-12
Auto-add tags to in-depth.html

camillobruni:
Use the benchmark definitions from JetStreamDriver.js to add tags to each workload title.

- Add tags
- Add links to run workload matching a tag
- Add links to run a single workload
- Improve in-depth.html css a bit to use less spacing around headers and the first definition

### DIFF
--- a/in-depth.html
+++ b/in-depth.html
@@ -139,9 +139,18 @@
             </h3>
 
             <p>
+<<<<<<< HEAD
                 The <span style="color: var(--text-color-very-subtle);">greyed-out workloads</span> are not run by default but can be manually enabled on the command-line or via
                 the <code>?test=</code> URL parameter.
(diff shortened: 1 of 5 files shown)


in-depth.html
```diff
@@ -885,8 +882,40 @@ <h3>
...
+    <script src="utils/params.js"></script>
+    <script src="JetStreamDriver.js"></script>
+    <script>
+        (function () {
+            for (const benchmark of BENCHMARKS) {
+                let dt = document.getElementById(benchmark.name);
+                if (!dt) continue;
+
+                if (benchmark.tags.has("disabled"))
+                    dt.classList.add("non-default");
```
danleh: Right now, this is done statically in the html, do you want to replace that completely with this dynamic code? Right now, this is superfluous.

PR #283, opened 2026-02-19
Partially address babylonjs leak

camillobruni:
We don't dispose of the babylon scene and engine which means we leak almost 100MB of cached data per iteration.

### DIFF
--- a/babylonjs/benchmark/scene.js
+++ b/babylonjs/benchmark/scene.js
@@ -55,7 +55,7 @@ class Benchmark {
   }
 
   async runIteration() {
-    const {classNames, cameraRotationLength, dispose: disposer} = await BabylonJSBenchmark.runComplexScene(
+    const {classNames, cameraRotationLength, disposer} = await BabylonJSBenchmark.runComplexScene(
       this.preloaded.fortData,
       this.preloaded.cannonData,
       this.preloaded.particlesJson,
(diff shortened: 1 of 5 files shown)


camillobruni:
As per discussions – too risky to land for v3.0, let's only land it on the main branch.

babylonjs/benchmark/scene.js
```diff
@@ -76,4 +78,10 @@ class Benchmark {
     if (value != expected)
       throw new Error(`Expected ${name} to be ${expected}, but got ${value}`);
   }
+
+  postIteration() {
+    // FIXME: move to measured phase. 
+    this.sceneDisposer();
+  }
```
kmiller68, line 85: Since we're not doing this in JS3 should we move this to runIteration?

babylonjs/benchmark/startup.js
```diff
@@ -80,4 +82,9 @@ class Benchmark extends StartupBenchmark {
     if (value != expected)
       throw new Error(`Expected ${name} to be ${expected}, but got ${value}`);
   }
+
+  postIteration() {
+    // FIXME: move to measured phase.
+    this.sceneDisposer();
+  }
```
kmiller68, line 89: Ditto.

kmiller68 (approved):
LGTM with comment.

PR #285, opened 2026-02-25
Prepare in-depth.html work (backmerge of #274)

camillobruni:
Backmerge of #274 (ad5e39771904b63750ae410fb00b71c9d2992b03):

- Format in-depth.html
- Add empty in-depth.html sections for the new workloads
- Add unit tests to check that all benchmarks have a matching in-depth entry

First steps to address #199.

### DIFF
--- a/in-depth.html
+++ b/in-depth.html
@@ -24,6 +24,7 @@
 -->
 <!DOCTYPE html>
 <html>
+
 <head>
     <meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


danleh (approved):
LGTM, looks the same as #274, as it should. Since we already discussed this in the meeting last week, I think it's fine if you just merge this to 3.0.

PR #288, opened 2026-02-26
Auto-add tags to in-depth.html

camillobruni:
Use the benchmark definitions from JetStreamDriver.js to add tags to each workload title.

- Add tags
- Add links to run workload matching a tag
- Add links to run a single workload
- Improve in-depth.html css a bit to use less spacing around headers and the first definition
- Use relative link from index.html => in-depth.html 

Reland of #279

### DIFF
--- a/in-depth.html
+++ b/in-depth.html
@@ -139,18 +139,9 @@
             </h3>
 
             <p>
-<<<<<<< HEAD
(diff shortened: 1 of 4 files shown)


danleh (approved):
Generally LGTM with one question. I would also wait for the deployment to be up, so we can inspect this once more before landing.

danleh (approved):
Still LGTM, thanks!

PR #291, opened 2026-03-05
Skip animations with non-default parameters for v3.0

camillobruni:
We don't want CSS animation to interfere with short running measurements which
can easaily happen if you only run a single workload.


danleh (approved):
LGTM, but wouldn't we also want to disable animations for the full suite via an additional parameter?

camillobruni:
For now this is probably good enough since you can disable it with "&startAutomtically" which we can use in our infra.
I'll finish the more involved approach in #200 where we can just wait for the animation to finish in all other cases.

danleh (approved):
Still LGTM, but let's have a look by Keith whether this would be fine for 3.0.

index.html
```diff
@@ -89,7 +89,12 @@
     <script src="JetStreamDriver.js"></script>
 
 </head>
-<body onload="initialize()">
+<body onload="initialize()" class="animate">
+<script>
+    if (!JetStreamParams.isDefault) {
+        document.body.classList.remove("animate");
```
kmiller68: It might be nice to set the non-default color CSS classes here so all the CSS changes for non-default happen in one place. Can you add a FIXME? Or is there some reason that wouldn't work?

kmiller68 (approved):
LGTM, have you had a chance to confirm on your end that it doesn't change the scores? My local testing of Safari in our harness was neutral.

camillobruni:
On chrome running on the local server I see no diff with the default params:
```
Before                               196.1  ± 1.1%
After                                196.18 ± 0.48%
```

index.html
camillobruni: 👍 Moved down the script to after the table and called the showNonDefaultParams from there.

PR #293, opened 2026-03-17
Fix typo for --test parsing

joyeecheung:
Fix typo for --test parsing

cc @camillobruni

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -119,7 +119,7 @@ const printHelp = cliParams.delete("help");
 const dumpTestList = cliParams.delete("dumpTestList");
 
 if (cliArgs.length) {
-    let tests = cliParams.has("test") ? cliParams.get("tests").split(",") : []
+    let tests = cliParams.has("test") ? cliParams.get("test").split(",") : []
     tests = tests.concat(cliArgs);
     cliParams.set("test", tests.join(","));
 }
--- a/tests/run-shell.mjs
+++ b/tests/run-shell.mjs
@@ -82,6 +82,13 @@ const TESTS = [
         },
     },
     {
+        name: "Test Flag With Positional Args",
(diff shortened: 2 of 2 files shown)


utils/shell-config.js
```diff
@@ -55,3 +55,8 @@ if (typeof performance == "undefined")
 
 performance.mark ??= function(){};
 performance.measure ??= function(){};
+
+if (isD8) {
+    performance.mark = function(){};
```
camillobruni: Mark and measure is supported in newer d8 versions (we do run CI tests so this does work). What d8 version were you using?

See https://crrev.com/c/6634602

cli.js
```diff
@@ -119,7 +119,7 @@ const printHelp = cliParams.delete("help");
 const dumpTestList = cliParams.delete("dumpTestList");
 
 if (cliArgs.length) {
-    let tests = cliParams.has("test") ? cliParams.get("tests").split(",") : []
+    let tests = cliParams.has("test") ? cliParams.get("test").split(",") : []
```
camillobruni, line 122: Could you add a shell test as well to https://github.com/WebKit/JetStream/blob/main/tests/run-shell.mjs so we don't accidentally regress this again?

utils/shell-config.js
joyeecheung: I was using a main branch pulled yesterday (or at least I believe so) - now I am running it again from a different computer and this no longer seems needed after I fixed the typo 🤯 some what puzzled about what's going on, but I think this can be removed..

cli.js
joyeecheung: Done - I now realize this only reproduces when there's a positional argument

camillobruni (approved):

PR #297, opened 2026-03-25
Use explicit 3.0 version in logo

camillobruni:
This makes it easier to spot minor fixes in the future.

- Add resources/JetStreamLogo.src.svg as editable svg
- Exported new logo to resources/JetStream30Logo.svg


resources/JetStreamLogo.src.svg
danleh, line 1: tiny typography nit: Could you very slightly reduce the letter spacing around the `.` in `3.0`?

danleh (approved):
LGTM with nit. Lets also update the main branch afterwards (this PR goes to the 3.0 branch directly)

kmiller68:
I don't have super strong feelings about this but one downside is that incremental updates require touching up the logo. I kinda prefer the paragraph at the top so we can quickly roll out the number.

camillobruni:
It would be in sync with speedometer where we did the same change (though the "logo" is just text):
- https://browserbench.org/Speedometer3.1/
- https://browserbench.org/Speedometer3.0/

It would just eliminate an easy source of error if we ever do an incremental dot-release.

kmiller68:
> It would just eliminate an easy source of error if we ever do an incremental dot-release.

What easy source of error are you thinking? That we copy the wrong branch?

I guess I'd prefer to make this change in 3.1 (if needed).

camillobruni:
Mostly about accidentally using the wrong minor version without noticing since the log doesn't feature the number immediately. It's a bit less of an issue with jetstream since it does mention the minor version in the text.

In the past these two were pretty easy to confuse (since it's only in the tab title).
- https://browserbench.org/Speedometer2.0/
- https://browserbench.org/Speedometer2.1/

PR #301, opened 2026-03-31
Workaround firewall blocking on browserbench.org

kmiller68:
While deploying JetStream3 on browserbench.org a bunch of the assets couldn't be loaded due to firewall restrictions. This change upstreams the hotfixes back to the main repo.

These were originally two different changes. I combined them into one for this PR.


camillobruni (approved):

camillobruni:
let's land this as is for the v3.0 branch – we should probably come up with a simple unittest for the main branch to prevent this in the future.

PR #304, opened 2026-04-22
Rewrite line number computation in WSL

### LINKED ISSUE #298: Rewrite line number computation in WSL
In Lexer.js, WSL [uses the following code to find the line number of an index into the source string](https://github.com/WebKit/JetStream/blob/de88e36ae91d5bd13126fa4cc4b0e0346d779842/WSL/Lexer.js#L54-L58):
 ```
   lineNumberForIndex(index)
    {
        let matches = this._text.substring(0, index).match(/\n/g);
        return (matches ? matches.length : 0) + this._lineNumberOffset;
    }
```
lineNumberForIndex is called more than 40000 times over the course of the benchmark. This approach is linear in the length of the source string, making the overall cost of lineNumberForIndex quadratic. 

In SpiderMonkey, this function is ~15% of the entire benchmark. [V8](https://chromium-review.googlesource.com/c/v8/v8/+/5886727) and [JSC](https://github.com/WebKit/WebKit/pull/34805) have each added "a caching mechanism to optimize repeated matching of global atomic regular expressions on growing substrings". This is a significant win on this workload, but is quite unlikely to be relevant in real-world code. AFAICT, WSL was written specifically for JetStream 2, and is not based on an existing framework. Unless I'm missing a reason that we should expect this cache to be useful in the wild, we should rewrite this code.

### PR DESCRIPTION
**Pull request by jrmuizel**
Instead of computing the line number on demand, keep track of it as we lex.

Previously, we ended up spending a quadratic amount of time computing the line number because we'd do something like:
```
for every index:
        count the number line breaks up to this index.
```

Fixes #298

### DIFF
--- a/WSL/Lexer.js
+++ b/WSL/Lexer.js
@@ -35,11 +35,12 @@ class Lexer {
         this._text = text;
         this._index = 0;
         this._stack = [];
+        this._lineNumber = lineNumberOffset;
     }
     
     get lineNumber()
     {
-        return this.lineNumberForIndex(this._index);
+        return this._lineNumber;
     }
     
     get origin() { return this._origin; }
@@ -51,17 +52,22 @@ class Lexer {
     
     get originKind() { return this._originKind; }
     
-    lineNumberForIndex(index)
+    _countNewlines(str, length)
     {
-        let matches = this._text.substring(0, index).match(/\n/g);
-        return (matches ? matches.length : 0) + this._lineNumberOffset;
+        let count = 0;
+        for (let i = 0; i < length; i++) {
+            if (str[i] === '\n')
+                count++;
+        }
+        return count;
     }
     
-    get state() { return {index: this._index, stack: this._stack.concat()}; }
+    get state() { return {index: this._index, stack: this._stack.concat(), lineNumber: this._lineNumber}; }
     set state(value)
     {
         this._index = value.index;
         this._stack = value.stack;
+        this._lineNumber = value.lineNumber;
     }
     
     static _textIsIdentifierImpl(text)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jrmuizel** [comment]
@camillobruni review ping

File: WSL/Lexer.js [hunk 1]
```diff
@@ -51,17 +52,22 @@ class Lexer {
...
+    _countNewlines(str, length)
     {
-        let matches = this._text.substring(0, index).match(/\n/g);
-        return (matches ? matches.length : 0) + this._lineNumberOffset;
+        let count = 0;
+        for (let i = 0; i < length; i++) {
+            if (str[i] === '\n')
+                count++;
+        }
+        return count;
```
kmiller68: I think a regex is the more idiomatic way to do this.
```suggestion
    let matches = str.substring(0, length).match(/\n/g);
    return matches ? matches.length : 0;
```

kmiller68 (changes requested):
Overall seems like a reasonable change. I do think most devs will end up using a regex for this so I'd mildly prefer keeping that. That's also how the original test was structured so all other things equal I'd prefer to keep it. 

I know it's maybe(?) faster to use the manual loop due to Array allocation but we shouldn't be optimizing the benchmark itself.

kmiller68 (approved):
LGTM

WSL/Lexer.js
```diff
@@ -51,17 +52,18 @@ class Lexer {
     
     get originKind() { return this._originKind; }
     
-    lineNumberForIndex(index)
+    _countNewlines(str, length)
     {
-        let matches = this._text.substring(0, index).match(/\n/g);
-        return (matches ? matches.length : 0) + this._lineNumberOffset;
+        let matches = str.substring(0, length).match(/\n/g);
+        return matches ? matches.length : 0;
```
camillobruni: nit: maybe modernize this as flyby too? `matches?.length ?? 0`

jrmuizel: Done.

jrmuizel:
@camillobruni can this be merged?

PR #307, opened 2026-05-06
Fix update-ui performance.measure nesting

camillobruni:
The mark and measure calls were out of sync and caused unclosed scopes.

### DIFF
--- a/JetStreamDriver.js
+++ b/JetStreamDriver.js
@@ -347,7 +347,8 @@ class Driver {
         performance.mark("update-ui-start");
         const start = performance.now();
         for (const benchmark of this.benchmarks) {
-            await benchmark.updateUIBeforeRun();
+            performance.mark("update-ui-start");
+            benchmark.updateUIBeforeRun();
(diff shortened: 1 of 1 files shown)


JetStreamDriver.js
```diff
@@ -344,9 +344,9 @@ class Driver {
         } else if (!JetStreamParams.dumpJSONResults)
             console.log("Starting JetStream3");
 
-        performance.mark("update-ui-start");
         const start = performance.now();
         for (const benchmark of this.benchmarks) {
+            performance.mark("update-ui-start");
             await benchmark.updateUIBeforeRun();
```
danleh: Since `updateUIBeforeRun` is not `async`, why the `await` here? Also note that the call to `benchmark.updateUIAfterRun()` is not using `await`.

danleh (approved):
LGTM with nit/question (somewhat orthogonal to the change, but maybe we can drive-by address this). Sorry for the late review, was OOO.

PR #312, opened 2026-08-01
Stop JS3 from silently hanging if simd is not supported

justinmichaud:
Stop JS3 from silently hanging if simd is not supported. Pipe through compilation error promise rejections so that they are not silently ignored.

In particular, this is useful for 32-bit jsc, which does not support wasm simd. Today, JS3 will just hang in the middle of a test run.


wasm/tfjs-bundle.js
```diff
@@ -23220,7 +23220,7 @@
                                             console.warn("Fatal error: no binary file found for ./wasm/tfjs-backend-wasm-simd.wasm and ./wasm/tfjs-backend-wasm.wasm");
                                         WebAssembly.instantiate(blob, info).then(function (output) {
                                             receiveInstance(output.instance, output.module);
-                                        });
+                                        }, readyPromiseReject);
```
kmiller68, line 23223: Is this needed won't this get picked up by the unhandled promise rejection event?

kmiller68 (approved):

PR #318, opened 2026-09-20
Reorganize workloads into a top-level workloads/ directory

### LINKED ISSUE #276: Cleanup workload directory structure
Definitely not urgent and only for after the 3.0 release, but it's a bit of a mess right now: Some workloads don't have their own directory (e.g., `tfjs-wasm` is just inside the `wasm/` directory); some Wasm workloads are under `wasm/`, some are not; and for those workloads in their own directory, the directory doesn't always match the workload name.

Let's:
- put all workloads under `workloads/`,
- put each workload in their own subdirectory, where the directory name matches precisely the workload name (except for those workloads that are closely related or have lots of files in common)
- don't put Wasm workloads under `wasm/` but just next to the JS ones.

harini9803:
Fixes #276.

Pure cleanup/reorg — no workload was changed materially.

- All workloads now live under `workloads/`.
- Each workload has its own subdirectory named exactly after its registered
  name, except closely-related workloads that share files (ARES-6, Octane,
  RexBench, SeaMonster, SunSpider, bigint, proxy, code-load, babylonjs,
  prismjs, transformersjs, Dart, dotnet, tfjs-wasm, intl,
  web-tooling-benchmark).
- The `wasm/` grouping is gone; Wasm workloads sit next to the JS ones.
- `simple/`, `generators/`, `class-fields/` and `worker/` were split, since
  their workloads shared no files.

912 of the 926 changed files are pure renames. Content edits are limited to
five root files (`JetStreamDriver.js`, `in-depth.html`, `.prettierignore`,
`.gitignore`, `tests/run-license.mjs`) and nine workload files: four
`webpack.config.mjs` (`../utils` -> `../../utils`), three sources with
hardcoded root-relative paths (below), plus a stale comment in
`argon2-wasm/benchmark.js` and a stale link in `dotnet/README.md`.

### Testing

- `test:chrome -- default` — 77 benchmarks, 0 failures
- `test:chrome -- disabled` — 17 benchmarks, 0 failures
- `test:in-depth` — 4/4, incl. the relative-link check
- `test:license` — passes; verified no files silently dropped from checking
- `prettier` — unchanged (same pre-existing warnings as `main`)
- V8 shell (jsvu v15.6.21): unit tests, CLI, `proxy-mobx`, grouped `SunSpider`,
  and `argon2-wasm` — the last one exercises the compressed-resource path that
  loads `zlib-wasm/shell.js`.

Firefox/Safari, and the jsc/spidermonkey shells, are left to CI.

### Notes for review

- Besides the mechanical path updates, three workload sources had root-relative
  paths hardcoded *inside* them — invisible to the driver's `files`/`preload`
  lists: `transformersjs/benchmark.js` (`localModelPath` and the ONNX `mjs`
  prefix), `tfjs-wasm/tfjs.js` (`WASM_TFJS_DIR`), and `zlib-wasm/shell.js`
  (shell-only zlib polyfill).
- `test:build` triggers on changed directories, and every workload directory
  moved, so CI will rebuild all 8 buildable workloads — expect a long run.
- `wasm/tsf.js` / `wasm/tsf.wasm` are unreferenced leftovers from the initial
  WebKit import, superseded by `tsf-wasm/build/`. I moved rather than deleted
  them — happy to drop them in a follow-up if you'd prefer.
- `RexBench/sunspider_benchmark.js` references `SunSpider/...`; it's a legacy
  standalone harness not registered in `BENCHMARKS`, and that path was already
  broken before this change, so I left it alone.


harini9803:
@danleh Pushed a fix for the Test Build failure - web-ssr/webpack.config.mjs referenced ../utils/BabelCacheBuster.mjs, which needed to become ../../utils/ after the move. Verified the build locally. Needs your approval again to run the workflows.

